PR #21, opened 2017-12-18
Add support for FileField response.

h-hirokawa:
DRF's serializers.FileField returns URI or the name of the field in response.
I added the support for it, so please check it.

PS Thank you for the package I have really wanted!

### DIFF
--- a/src/drf_yasg/utils.py
+++ b/src/drf_yasg/utils.py
@@ -290,10 +290,10 @@ def serializer_field_to_swagger(field, swagger_object_type, definitions=None, **
         err = SwaggerGenerationError("parameter of type file is supported only in a formData Parameter")
         if swagger_object_type == openapi.Schema:
             # FileField.to_representation returns URL or file name
+            result = SwaggerType(type=openapi.TYPE_STRING, read_only=True)
             if getattr(field, 'use_url', api_settings.UPLOADED_FILES_USE_URL):
-                return SwaggerType(type=openapi.TYPE_STRING, format=openapi.FORMAT_URI)
-            else:
-                return SwaggerType(type=openapi.TYPE_STRING)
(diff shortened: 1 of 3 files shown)


src/drf_yasg/utils.py
```diff
@@ -287,7 +288,13 @@ def make_schema_definition():
         # swagger 2.0 does not support specifics about file fields, so ImageFile gets no special treatment
         # OpenAPI 3.0 does support it, so a future implementation could handle this better
         err = SwaggerGenerationError("parameter of type file is supported only in a formData Parameter")
-        if swagger_object_type != openapi.Parameter:
+        if swagger_object_type == openapi.Schema:
+            # FileField.to_representation returns URL or file name
+            if getattr(field, 'use_url', api_settings.UPLOADED_FILES_USE_URL):
+                return SwaggerType(type=openapi.TYPE_STRING, format=openapi.FORMAT_URI)
```
axnsan12: Since this only applies to response schemas, would it make sense to always return a `read_only` Schema here?

axnsan12: I just realized that doing that would involve moving some code around (from [here](https://github.com/axnsan12/drf-yasg/blob/a2c21539f783480e99177c64b0fb79bd61f3e318/src/drf_yasg/utils.py#L216) to [here](https://github.com/axnsan12/drf-yasg/blob/a2c21539f783480e99177c64b0fb79bd61f3e318/src/drf_yasg/utils.py#L179)) and adding some checks to prevent the value on the field overriding the value passed as argument; if you don't to deal with that I could handle it.

axnsan12:
👍 I'm going to rebase this on #20 and merge it

h-hirokawa:
Thank you for merging!

PR #23, opened 2017-12-20
Pass context when instanciating a serializer

ticosax:
Hi,
Thank you for this project, I'm very impressed.

Some of our serializers needs the request to render properly, this change aim to fix that.

The patch is inspired by DRF itself.
https://github.com/encode/django-rest-framework/blob/master/rest_framework/generics.py#L111

### DIFF
--- a/src/drf_yasg/inspectors.py
+++ b/src/drf_yasg/inspectors.py
@@ -114,7 +114,8 @@ class SwaggerAutoSchema(object):
         """
         if not hasattr(self.view, 'get_serializer'):
             return None
-        return self.view.get_serializer()
+        kwargs = {'context': self.view.get_serializer_context()}
+        return self.view.get_serializer(**kwargs)
 
     def get_request_serializer(self):
         """Return the request serializer (used for parsing the request payload) for this endpoint.


axnsan12:
Hello and thank you!

Looks good and makes sense at first glance.

Could you provide some examples for this usage? Perhaps even some tests would be useful to make sure this keeps working in the future.

ticosax:
sure thing. I will add the test.
the time being, here is the link explaining why it might comes handy.
http://www.django-rest-framework.org/api-guide/serializers/#including-extra-context

src/drf_yasg/inspectors.py
```diff
@@ -114,7 +114,8 @@ def get_view_serializer(self):
         """
         if not hasattr(self.view, 'get_serializer'):
             return None
-        return self.view.get_serializer()
```
axnsan12: I don't follow, though. In [the link you sent](https://github.com/encode/django-rest-framework/blob/a81e60ff390b9b6788b2cac24a01ee4dd2dcffd2/rest_framework/generics.py#L111), you can see that `get_serializer` already adds the serializer context before instantiating `serializer_class`.

src/drf_yasg/inspectors.py
```diff
@@ -114,7 +114,8 @@ def get_view_serializer(self):
...
+        kwargs = {'context': self.view.get_serializer_context()}
```
axnsan12: In fact, this line is exactly like the line you linked to, and it is in the same function that is called above.

axnsan12:
I think the real problem you are encountering is not that context is not added to the serialzier, but that `context['request']` is `None`. Unfortunately, this is just the way it is, there is no request when the schema generator inspects views, because they are artificially instantiated.

What do you mean that your serializer "needs the request to render properly"? Could you provide an example of such a serializer?

src/drf_yasg/inspectors.py
ticosax: damn, you are right!

ticosax:
request is None when the generator is configured with `public=True`

https://github.com/axnsan12/drf-yasg/blob/master/src/drf_yasg/generators.py#L85

That's in fact enough to have the request.
Thank you.

axnsan12:
Do note that this request will be, for all views, the request that arrived **against the schema view** (e.g. `GET /swagger.json`) and not a request that would normally hit your serializer, with query parameters, different HTTP methods or whatnot.

ticosax:
Sure, as a side effect we have a different documentation rendered, depending if user is connected or not and so on.
Our permission policy is reflected on the documentation that way.

axnsan12:
In 1.0.6, views should now always have the schema view request bound, regardless of `public=True`.

PR #24, opened 2017-12-20
allow CurrentUserDefault() to be validated and serialized.

ticosax:
This default value needs to receive special attention to be validated and serialized.

### DIFF
--- a/src/drf_yasg/codecs.py
+++ b/src/drf_yasg/codecs.py
@@ -4,6 +4,7 @@ from collections import OrderedDict
 
 from coreapi.compat import force_bytes
 from future.utils import raise_from
+from rest_framework.serializers import CurrentUserDefault
 from ruamel import yaml
 
 from . import openapi
--- a/src/drf_yasg/codecs.py
+++ b/src/drf_yasg/codecs.py
@@ -4,7 +4,6 @@ from collections import OrderedDict
 
 from coreapi.compat import force_bytes
 from future.utils import raise_from
-from rest_framework.serializers import CurrentUserDefault
 from ruamel import yaml
 
 from . import openapi
--- a/src/drf_yasg/utils.py
+++ b/src/drf_yasg/utils.py
@@ -182,9 +182,8 @@ def serializer_field_to_swagger(field, swagger_object_type, definitions=None, **
         if swagger_object_type != openapi.Items and 'default' not in instance_kwargs:
             default = getattr(field, 'default', serializers.empty)
             if default is not serializers.empty:
-                if callable(default) and hasattr(default, 'set_context'):
-                    default = str(default)
-                instance_kwargs['default'] = default
+                if not (callable(default) and hasattr(default, 'set_context')):
+                    instance_kwargs['default'] = default
         if swagger_object_type == openapi.Schema and 'read_only' not in instance_kwargs:
             if field.read_only:
                 instance_kwargs['read_only'] = True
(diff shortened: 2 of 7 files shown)


src/drf_yasg/utils.py
```diff
@@ -182,6 +183,8 @@ def SwaggerType(**instance_kwargs):
         if swagger_object_type != openapi.Items and 'default' not in instance_kwargs:
             default = getattr(field, 'default', serializers.empty)
             if default is not serializers.empty:
+                if isinstance(default, CurrentUserDefault):
```
ticosax: reviewing myself again, I'm not sure about the fix. Maybe I should do like DRF is doing...
https://github.com/encode/django-rest-framework/blob/master/rest_framework/fields.py#L253
yeah, seems better.

src/drf_yasg/codecs.py
```diff
@@ -4,6 +4,7 @@
 
 from coreapi.compat import force_bytes
 from future.utils import raise_from
+from rest_framework.serializers import CurrentUserDefault
```
vishes-shell: Unused import

tests/reference.yaml
```diff
@@ -1054,6 +1054,9 @@ definitions:
         type: string
         format: date
         readOnly: true
+      owner:
+        type: string
+        default: CurrentUserDefault()
```
axnsan12: I don't see how this is appropriate. 

The string `CurrentUserDefault()` does not represent a valid user ID, so cannot be a "default value" for the owner field which is a PrimaryKeyRelatedField.

As far as I know, the OpenAPI specification provides no way to specify a default based on external state such as currently authenticated user, so this can't really be expressed by spec. The best you could hope for would be to document the behavior in a description, e.g. "if `owner` is not provided, it will be assumed to be the currently logged in user".

Since this project aims for spec compliance and codegen tool compatibility, we cannot just use arbitrary strings just for documentation purposes in the `default` field.

axnsan12: In fact, if [this TODO](https://github.com/axnsan12/drf-yasg/blob/7683a28816f85478fda96a60eb2fa2fe10bc4c38/src/drf_yasg/utils.py#L242) were to be implemented, the owner field would be inferred to be of type `int` (like `pk` of `User`), and "CurrentUserDefault()" would not even conform to the type.

According to [the OpenAPI spec](https://github.com/axnsan12/drf-yasg/blob/7683a28816f85478fda96a60eb2fa2fe10bc4c38/src/drf_yasg/utils.py#L242), the value of `default` must conform to the `Schema` it is declared in.

axnsan12:
I would say your best option in a situation like this would be to add a `help_text` to the `owner` field describing what happens in the default case. The treatment proposed here doesn't seem like a good idea to me.

ticosax:
I have no idea about what I'm doing here. I haven't study in depth openapi/swagger specs.I completely trust your guidance.
Anyway, writing the regression test was the hardest part.
I'll try to make it better following your comments.

axnsan12:
I will try to fix this so that it will at least not generate an invalid schema (see #25). I think I have a solution in mind which would be able to show the user ID of the currently logged in user that is viewing the schema for CurrentUserDefault. That wouldn't be perfect, but at least it's something, and together with the description suggested above would be decent I guess.

axnsan12:
Hello,

Can you confirm if 1.0.6 fixes your problem?

ticosax:
I confirm it works well. Thank you.

PR #26, opened 2017-12-21
Adjust type for PrimaryKeyRelatedField

ticosax:


src/drf_yasg/utils.py
```diff
@@ -238,8 +238,9 @@ def make_schema_definition():
             items=child_schema,
             unique_items=True,  # is this OK?
         )
+    elif isinstance(field, serializers.PrimaryKeyRelatedField):
+        return SwaggerType(type=openapi.TYPE_INTEGER)
```
axnsan12: This is not right, a Model's primary key can have a lot more different types than `int`. You would have to check the actual `queryset` on the field to check the `ModelField` details. 

For example, the primary key could be a string, or a datetime, or it could be a compound key. There is a semi-attempt at guessing type from the Model's `pk` in `generators.py` in the function that inspects path parameters, but it's not perfect/complete there either.

axnsan12:
Seems good, thanks!

PR #71, opened 2018-02-26
Make 1.10.0 the min six version

bryant1410:
I believe some projects could have transitive dependency conflicts with only supporting the latest version of six. So, what about supporting the previous version? I tested it locally and tests passed on some python versions. I think they'll all pass.


axnsan12:
Ok, no problem, I'll merge if all tests pass :)

PR #79, opened 2018-03-09
Added HiddenFieldInspector (#78)

### LINKED ISSUE #78: Don't display fields of type HiddenField
Hi,   It seems that drf-yasg doesn't support hiding serializers.HiddenField's yet?  A hidden field in my serializer is showing up in the example data and model docs.

It looks like this is a known issue due to this TODO - https://github.com/axnsan12/drf-yasg/blob/9ad55bac9931a93ba2a9f7fd9f8693a0f8070c38/src/drf_yasg/inspectors/field.py#L472 but raising it.

therefromhere:
Attempt to resolve #78 , is this on the right track?  If so I'll look into how to add tests for this.

### DIFF
--- a/src/drf_yasg/inspectors/field.py
+++ b/src/drf_yasg/inspectors/field.py
@@ -465,11 +465,21 @@ class DictFieldInspector(FieldInspector):
         return NotHandled
 
 
+class HiddenFieldInspector(FieldInspector):
+    """Hide ``HiddenField``."""
+
+    def field_to_swagger_object(self, field, swagger_object_type, use_references, **kwargs):
+        if isinstance(field, serializers.HiddenField):
+            return None
+        else:
+            return NotHandled
+
+
 class StringDefaultFieldInspector(FieldInspector):
     """For otherwise unhandled fields, return them as plain :data:`.TYPE_STRING` objects."""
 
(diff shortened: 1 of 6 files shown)


axnsan12:
The implementation looks about right, but it could use some simple tests (basically just add a HiddenField somewhere in ``testproj`` and regenerate ``reference.yaml`` - https://github.com/axnsan12/drf-yasg/blob/master/CONTRIBUTING.rst#pull-requests).

The docs for ``DEFAULT_FIELD_INSPECTORS`` should also be updated to match its new reality.

src/drf_yasg/inspectors/field.py
```diff
@@ -465,11 +465,21 @@ def field_to_swagger_object(self, field, swagger_object_type, use_references, **
         return NotHandled
 
 
+class HiddenFieldInspector(FieldInspector):
+    """Hide ``HiddenField``."""
+
+    def field_to_swagger_object(self, field, swagger_object_type, use_references, **kwargs):
+        if isinstance(field, serializers.HiddenField):
+            return None
+        else:
```
axnsan12: The ``else`` here is not needed, and could be removed to keep the style consistent with the other classes above.

therefromhere:
done

axnsan12:
Seems ok 👍, thanks!

PR #80, opened 2018-03-09
Fix URL in the example

ko-pp:
Django's URLs are regexes, thus the character '.' is used as a joker, making `r'^swagger(?P<format>.json|.yaml)$'` accept not only `swagger.json` and `swagger.xml` but any URL replacing the '.' with a single other character, leading to a TypeError: ```Expected a `openapi.Swagger` instance```.
Simply escape this character to fix the "issue"

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -134,7 +134,7 @@ In ``urls.py``:
    )
 
    urlpatterns = [
-      url(r'^swagger(?P<format>.json|.yaml)$', schema_view.without_ui(cache_timeout=None), name='schema-json'),
+      url(r'^swagger(?P<format>\.json|\.yaml)$', schema_view.without_ui(cache_timeout=None), name='schema-json'),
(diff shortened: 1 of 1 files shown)


axnsan12:
👍

nicholasgcoles:
Hello,

I seem to get this error every time I load my swagger docs (hosted on a server). 
`TypeError: Expected a openapi.Swagger instance`

But then after a couple seconds it loads everything perfectly.

Unable to replicate it locally for some reason.

Would there be a way for me to somehow disable this check?

ko-pp:
Hello,

Your error doesn't seem related to this PR, you should open a new issue and try to give additional information (for instance, a complete traceback of the exception).

Being able to replicate the error with a minimal project would be huge help to find the problem.

Lastly, a `TypeError` most likely does not indicate a failed check but an error in the code and thus can't be disabled, only fixed by finding the root cause.

PR #81, opened 2018-03-09
Fix schema generation with OneToOneFields

ko-pp:
Having a OneToOneField breaks (AttributeError) generators.OpenAPISchemaGenerator.get_path_parameters() because neither model_field.help_text nor model_field.primary_key are defined so I added hasattr() to both if-statement to guard against the exception.

All the "description" keys of the generated schema are now empty so there might be a bigger issue here but since this patch makes drf-yasg work for my project, I open this PR nonetheless.

Tested with Django 1.11.10


src/drf_yasg/generators.py
```diff
@@ -409,9 +409,9 @@ def get_path_parameters(self, path, view_cls):
             if getattr(view_cls, 'lookup_field', None) == variable and attrs['type'] == openapi.TYPE_STRING:
                 attrs['pattern'] = getattr(view_cls, 'lookup_value_regex', attrs.get('pattern', None))
 
-            if model_field and model_field.help_text:
+            if model_field and hasattr(model_field, 'help_text') and model_field.help_text:
```
axnsan12: ``hasattr(model_field, 'help_text') and model_field.help_text`` could be replaced by ``getattr(model_field, 'help_text', '')``

src/drf_yasg/generators.py
```diff
@@ -409,9 +409,9 @@ def get_path_parameters(self, path, view_cls):
...
                 description = force_text(model_field.help_text)
-            elif model_field and model_field.primary_key:
+            elif model_field and hasattr(model_field, 'primary_key') and model_field.primary_key:
```
axnsan12: ...and this by ``getattr(model_field, 'primary_key', False)``

axnsan12:
A regression test for this would also be great, if you would like to add one (as described [https://github.com/axnsan12/drf-yasg/pull/79#issuecomment-371736246](here))

ko-pp:
Sorry for not thinking of getattr(), I'm pretty tired and didn't think beyond "does it work?"…
I'll have a go at trying to add a regression test as soon as I can but not before tomorrow unfortunately.
Anyway, thanks for this project :-)

axnsan12:
No problem and no rush! :smile:

axnsan12:
Nope, seems great. Thanks!

PR #106, opened 2018-04-22
Make parameters order stable for multi-parameter URLs

### LINKED ISSUE #105: URL with multiple parameters produces unstable doc ordering
If a URL has multiple parameters, currently the order of those isn't stable in the generated swagger doc (because the values are coming from a `set`).

This is causing intermittent failures in my swagger fixture test.

therefromhere:
Resolves #105 

Without the sorted() then the `parameters` ordering here is random depending on `PYTHONHASHSEED`, so test will fail ~50% of time:

```
    parameters:
      - name: todo_id
        in: path
        required: true
        type: string
      - name: yetanother_id
        in: path
        required: true
        type: string

```
https://github.com/axnsan12/drf-yasg/compare/master...therefromhere:stable_parameters_order?expand=1#diff-a9b50452eecdae91c8e1e478e092b2e4R561

It would be nice if we could show parameters in the same ordering as the URL, but unfortunately `uritemplate` doesn't provide that.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,6 +4,14 @@ Changelog
 
 
 *********
+**1.6.2**
+*********
+
+*Release date: TODO, 2018*
+
+- **FIXED:** made documentation ordering of parameters stable for urls with multiple parameters (:issue:`105`, :pr:`106`)
+
+*********
 **1.6.1**
 *********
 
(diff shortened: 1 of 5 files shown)


axnsan12:
Seems good, thanks 👍

PR #108, opened 2018-04-24
Be precise about configuration value types in Swagger UI settings docs

tuky:
Especially the
```
`None`: everything is collapsed
```
got me confused ;-)

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -190,8 +190,8 @@ OPERATIONS_SORTER
 Sorting order for the operation list of each tag.
 
 * :python:`None`: show in the order returned by the server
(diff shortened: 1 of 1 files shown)


vishes-shell:
@tuky why you've changed `None` to `'none'`?

tuky:
Simply put because `None` has no effect. It is also documented for `docExpansion` at https://github.com/swagger-api/swagger-ui/blob/112bca906553a937ac67adc2e500bdeed96d067b/docs/usage/configuration.md#parameters. It accepts `String=["list"*, "full", "none"]`.

axnsan12:
Yep, seems right, thanks! 👍

PR #112, opened 2018-04-27
Set min_length when allow_blank is false

elnappo:
As `allow_blank` is not specified in Swagger/OpenAPI 2.0 (see [here](https://swagger.io/docs/specification/data-models/data-types/#string)) this patch uses `min_length` to accomplish the same.


axnsan12:
The idea seems ok.

Can you try rebasing on master? https://github.com/axnsan12/drf-yasg/commit/209201b9a52826d3b8d617b571921cd6eb8ebda2 should fix your build errors

elnappo:
Done.

Issue #58, opened 2018-02-08
Hiding APIs that require authentication


lukik:
I've setup DRF-YASG but it seems not able to hide APIs that require Authentication.

Below is the configuration.

```
public_apis = [ 
    # sample-end-point
    url(f'{URL_VERSION}/sample-end-point/', include(sample_end_point.urls)),  # requires authentication
]

  schema_view = get_schema_view(
      openapi.Info(
        title="Swagger Doc",
        default_version='v1',
        description="Test description",
        terms_of_service="https://www.google.com/policies/terms/",
        contact=openapi.Contact(email="contact@snippets.local"),
        license=openapi.License(name="BSD License"),
      ),
      validators=['flex', 'ssv'],
      permission_classes=(permissions.AllowAny,), 
      public=False,
      patterns=public_apis,
  )

urlpatterns = public_apis + [

    url(r'^swagger(?P<format>.json|.yaml)$', schema_view.without_ui(cache_timeout=None), name='schema-json'),
    url(r'^swagger/$', schema_view.with_ui('swagger', cache_timeout=None), name='schema-swagger-ui'),
    url(r'^redoc/$', schema_view.with_ui('redoc', cache_timeout=None), name='schema-redoc'),
]
```
With the above configuration, the `sample-end-point` does not appear on the SwaggerUi. It only shows the `Authorize` Button and text that says `No operations defined in spec!`. But if I change the`public=False` to `public=True` setting, the `sample-end-point` appears. But that is not what I want as this URL should be accessible only after one logs in.

Is there something am not understanding or a setting am missing?

My Swagger settings are as below:

```
SWAGGER_SETTINGS = {
    'USE_SESSION_AUTH': False,
    'SECURITY_DEFINITIONS': {
        'api_key': {
            'type': 'apiKey',
            'in': 'header',
            'name': 'Authorization'
        }
    },
    'APIS_SORTER': 'alpha',
    'SUPPORTED_SUBMIT_METHODS': ['get', 'post', 'put', 'delete', 'patch'],
    'OPERATIONS_SORTER': 'alpha'
}
```

I'd posted the issue on [Stack Overflow](https://stackoverflow.com/questions/48687780/authenticating-swagger-api-docs-drf-yasg)

axnsan12:
Hello!

[...]

PR #130, opened 2018-05-22
apply fix from #58 to _SpecRenderer

terencehonles:
Same issue as #58 but a `TypeError` is thrown which causes a logged error and the `schema.{json|yaml}` is a 500 error rather than the appropriate 403 forbidden


axnsan12:
I think ``JSONRenderer`` would be more fitting here, seeing that this is a JSON/YAML endpoint generally. The idea makes sense though :+1:

terencehonles:
Thanks for incorporating your feedback!

PR #133, opened 2018-05-30
FileFieldInspector should not set read_only=True by default.

werwty:
The schema generated from a FileField with required=True set will
generate the following error when validated:
`Read only properties cannot be marked as required by a schema.`


axnsan12:
This is not the right approach. FileField is always ``read_only`` when represented as a ``Schema``, since you can't upload files via JSON, and the representation is just the file's url.

Instead, the loop [here](https://github.com/axnsan12/drf-yasg/blob/a4a11ad1ab0a594da7b5844126924466638db496/src/drf_yasg/inspectors/field.py#L70) should be modified to prevent ``read_only`` properties from being added into the ``required`` array, which would also be a more generic fix for the issue.

PR #142, opened 2018-06-11
swagger_extra_fields should be allowed even when we don't call get_schema

andrewyager:
It appears that it is possible to write a Schema Definition for use with a SchemaRef without creating the object through `get_schema`. The side effect of this is that if you have specified `schema_extra_fields` with parameters you care about (such as examples) they never get set.

This patch adds support for custom fields even when building a SchemaRef because it is possible for `field_to_swagger_object` to get called without calling `get_schema`. The below example illustrates a case where this happens.

Example:
```
class TestObj(object):
    foobar = None
    barfoo = None

    def __init__(self, **kwargs):
        setattr(self, 'foobar', kwargs.get('foobar'))
        setattr(self, 'barfoo', kwargs.get('barfoo'))


class TestSerializer(serializers.Serializer):
    barfoo = serializers.CharField(max_length=50)
    foobar = serializers.CharField(max_length=300)

    def create(self, data):
        return TestObj(**data)

    class Meta:
        swagger_schema_fields = {
            'example': {
                'barfoo': '1234',
                'foobar': '456'
            }
        }


class TestView(viewsets.ViewSet):
    """
    Test example view
    """

    permission_classes = [permissions.IsAuthenticated]
    lookup_field = 'address_partial'

    @swagger_auto_schema(
        operation_id='testview_get',
        description='#TestSerializer Example',
        responses={
            '200': TestSerializer(many=True)
        },
        permission_classes=[permissions.IsAuthenticated])
    def retrieve(self, request, pk=None, address_partial=None):
        response = {
            'hits': [
                {
                  'barfoo': 'asdf',
                  'foobar': 'foobar'
                }
            ]
        }
        serializer = TestSerializer(
            instance=response['hits'],
            many=True)
        return Response(serializer.data, status=status.HTTP_200_OK)
```

Without the patch, this will omit the `example` additional field. With the patch, it won't.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,6 +4,18 @@ Changelog
 
 
 *********
+**1.9.0**
+*********
+
+*Release date: Jun 16, 2018*
+
+- **ADDED:** added ``DEFAULT_GENERATOR_CLASS`` setting and ``--generator-clas`` argument to the ``generate_swagger``
+  management command (:issue:`140`)
+- **FIXED:** fixed wrongly required ``'count'`` response field on ``CursorPagination`` (:issue:`141`)
+- **FIXED:** fixed crash when encountering ``coreapi.Fields``\ s without a ``schema`` (:issue:`143`)
+
+
+*********
 **1.8.0**
 *********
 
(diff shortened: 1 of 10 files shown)


andrewyager:
(In my defence, the build was failing before this PR 😄)

PR #153, opened 2018-06-27
Allow override response serializer

xiaohanzhang:
Hi all,
Thanks for the great work. I was wondering if we can separate get_response_serializer and get_request_serializer ? So people can override them to get different schema in request and response.

Here is my use case:
```
class MyAutoSchema(SwaggerAutoSchema):
    def get_request_body_schema(self, serializer):
        return super().get_request_body_schema(serializer.get_request_serializer())

    def get_response_serializer(self):
        serializer = self.get_request_serializer() or self.get_view_serializer()
        return serializer.get_response_serializer()

class RequestSerializer(ModelSerializer):
    class Meta:
        model = Foo
        fields = (id, name, relation1, )

class ResponseSerializer(RequestSerializer):
    relation1 = NestedSerializer()

class MySerializer(ModelSerializer):
    def get_response_serializer(self):
        return ResponseSerializer(context=self.context)

    def get_request_serializer(self):
        return self
```


axnsan12:
Alright, seems fair.

PR #155, opened 2018-06-29
Handle duplicate urls in urlconf

jcushman:
Django resolves urls in order from top to bottom, and only uses the first matching URL found. This can happen for example when overriding one view in an existing app:

```
    path('accounts/login/', user_views.LoginView.as_view(), name='login'),
    path('accounts/', include('django.contrib.auth.urls')),  # logout, password change, password reset
```

drf-yasg doesn't follow Django's behavior, and instead uses the last view for a given path+method. This means duplicate URLs are documented incorrectly in the resulting schema.

This PR adds a test to show the correct behavior, and a filter to `OpenAPISchemaGenerator.get_api_endpoints` to ensure that each path is only included once.

### DIFF
--- a/src/drf_yasg/generators.py
+++ b/src/drf_yasg/generators.py
@@ -72,7 +72,7 @@ class EndpointEnumerator(_EndpointEnumerator):
 
         return path
 
-    def get_api_endpoints(self, patterns=None, prefix='', app_name=None, namespace=None):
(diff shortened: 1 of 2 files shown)


axnsan12:
Awesome, thanks for the catch!

jcushman:
Thanks for the test fix and quick merge!

PR #161, opened 2018-07-06
make generate_swagger work for projects without authentication

etene:
Hello,

I needed to generate the schema with `./manager.py generate_swagger` for a project on which there isn't any Django authentication configured.

The issue was that the `User` import in `generate_swagger.py` failed in this environment, so the schema wasn't generated.

I fixed it so that the schema can still generated. Passing the `--user` option also raises an error if auth is unconfigured. It passes the tests and the linter.

Could you consider merging ? That would be a very nice feature to have in the official repo for me (at least).

Thank you !

### DIFF
--- a/src/drf_yasg/management/commands/generate_swagger.py
+++ b/src/drf_yasg/management/commands/generate_swagger.py
@@ -4,7 +4,6 @@ import os
 from collections import OrderedDict
 from importlib import import_module
 
-from django.contrib.auth.models import User
 from django.core.exceptions import ImproperlyConfigured
 from django.core.management.base import BaseCommand
 from rest_framework.test import APIRequestFactory, force_authenticate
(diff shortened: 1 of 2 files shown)


etene:
Okay I don't really get in what way the ci failures are related to my changes, but if it's the case I'd be glad to fix :)

axnsan12:
Looking at the code again, we should probably use ``get_user_model`` instead of directly importing ``User``. But the idea of this patch is good, 

> Okay I don't really get in what way the ci failures are related to my changes, but if it's the case I'd be glad to fix :)

It seems unrelated to what you did, it's ok.

etene:
Thanks for your suggestion, I've made changes, is that better ?

axnsan12:
Yep, thanks!

etene:
Hello, can you tell me if you're planning on releasing soon ? I'm waiting for the next pypi release to have this feature integrated cleanly in a project I'm working on. Thank you !

etene:
I'll wait, no problem. Thank you for your time !

axnsan12:
In case you haven't noticed, this was released last Friday. Cheers!

PR #164, opened 2018-07-10
Schema is split to request and response parts

matllubos:


matllubos:
I think that for swagger 2.0 schema should be split to request and response because there is missing writeOnly value (added in 3.0). Result swagger file is invalid if serializer contains write_only or read_only fields

matllubos:
I will try to fix test in the next days

sobolevn:
+1 this is totally required.
@matllubos thanks!

axnsan12:
This is a bit invasive with adding arguments to all inspector methods. It's also not very robust - one might also want to generate different schemas according to request method (e.g. all fields optional for PATCH).

 The behaviour would also have to be behind a setting since not every use case might want this.

JoelLefkowitz:
@matllubos thanks for the time put into this PR. Could you bring the branch up to date with the 1.21.x branch or enable me to push to it?

PR #198, opened 2018-09-05
utils: add missing 'deprecated' and 'operation_summary' to swagger_auto_schema internal decorators data

swistakm:
This was the last missing piece to make it work. Without that the `summary` and `deprecated` did not appear in generated OpenAPI schema.


axnsan12:
Duplicate of #194.

Issue #41, opened 2018-01-11
Support django-rest-swagger style YAML docstrings


axnsan12:
See http://django-rest-swagger.readthedocs.io/en/stable-0.3.x/yaml.html

Sometimes it can be easier to specify overrides in docstrings rather than code - for example, complex `Schema`, `Parameter` or `Response` objects. It could be worth looking into also being able to supply `swagger_auto_schema` overrides as YAML in view docstrings.

Opa-:
This would be cool to handle extra `headers` parameters type 😃

PR #200, opened 2018-09-06
Allow description of parameters of ViewSets not using Models

PaulWay:
Allow parameters of ViewSets not using Models to describe their parameters in their docstring.


PaulWay:
I dunno why the Python 3.7 fetch failed, but that doesn't seem to be because of my code...

axnsan12:
See #201.

I would rather see a more general implementation of #41 rather than piecemeal chunks.
This PR also lacks required documentation and tests.

PR #217, opened 2018-09-20
Set x-nullable based on field informations

remilapeyre:
Many fields may be set a nullable in an API. While not covered explicitly
by Swagger 2, this information is usually indicated as a
[vendor extension](https://swagger.io/docs/specification/2-0/swagger-extensions/)
using the x-nullable field.

### DIFF
--- a/src/drf_yasg/inspectors/base.py
+++ b/src/drf_yasg/inspectors/base.py
@@ -192,7 +192,7 @@ class FieldInspector(BaseInspector):
 
             - arguments specified by the ``kwargs`` parameter of :meth:`._get_partial_types`
             - ``instance_kwargs`` passed to the constructor function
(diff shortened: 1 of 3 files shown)


axnsan12:
Hey - sorry for the delayed response.

This seems allright, it's pretty non-intrusive and seems to be accepted by many tools. 

However it seems to me that it should only apply to JSON Schemas, not path/query/form/header parameters, am I wrong?

remilapeyre:
Hi, thanks for the feedback.

I did not dive much in drf-yasg so I'm not sure what you are refering to.
Reading the docstring of `_get_partial_types` I thought my change would only be reflected on fields (`Helper method to extract generic information from a field`) so JSON Schemas and form data I think.

I do not know much of the usage of vendor extensions on Swagger but the resulting YAML seems fine to me.

PR #219, opened 2018-09-24
DjangoRestResponsePagination: fields description

gustabot42:
Description for pagination query parameters fields


axnsan12:
Hey! Thanks for the PR, but I don't like the idea of hardcoding display strings in the library.

PR #220, opened 2018-09-24
Meta docstring

gustabot42:
Using markdown metadata to get some id, summary, tags an deprecated.


axnsan12:
I don't really like this idea since it brings in support for yet another flavor of non-standard markdown, complicating both implementation and usage. It seems better suited in a project-specific schema class.

Issue #207, opened 2018-09-10
Remove the `.idea` folder from the repository


karimone:
Is the pycharm configuration, it simply should be not there.

axnsan12:
It is (mostly) non-user-specific PyCharm configuration, such as inspection settings, ignored directories, etc; it's helpful to have this data shared between everyone that uses PyCharm.

varnion:
I opened #237 before viewing this. I had a bad experience some yrs ago opening [a project](https://github.com/varnion/FastaDB) with these configs present. Maybe pinpointing which files are exclusive for user-specific configs and only ignoring them could help.

axnsan12:
> Maybe pinpointing which files are exclusive for user-specific configs and only ignoring them could help.

It's already like that, most user files should be ignored. Did you have any problems when opening the project?

PR #237, opened 2018-10-24
Ignore Pycharm config folder

varnion:


vishes-shell:
Not sure that this will be accepted since #207

varnion:
> Not sure that this will be accepted since #207

I see... thanks for the heads up!

axnsan12:
It will be accepted if there is a compelling reason 😄 But for me, I don't see any problems with the current set of included files, and it has helped me personally when working from multiple computers.

PR #258, opened 2018-11-29
Adding Exception language to let the developer know where the problem is

PauloPeres:
Adding a text to let the user know he forgot to add the filter_backends as an iterable object.
This way it's easier to know in what view the user forgot to do so

### DIFF
--- a/src/drf_yasg/inspectors/base.py
+++ b/src/drf_yasg/inspectors/base.py
@@ -335,8 +335,11 @@ class ViewInspector(BaseInspector):
             return []
 
         fields = []
-        for filter_backend in self.view.filter_backends:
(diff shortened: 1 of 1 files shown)


axnsan12:
This library is not a tool for debugging broken REST framework views. In general, if the view would not work as-is, we don't care about it.

Ideally any broken view would be silently skipped, at most logging a warning with the stack trace. It might make sense to do this with a catch-all try-except [here](https://github.com/axnsan12/drf-yasg/blob/f587785eb4b1830dd28eb9a815200c82ee7ca371/src/drf_yasg/generators.py#L355-L360).

PauloPeres:
You are Correct forget about this Pull Request

PR #259, opened 2018-11-30
Tags override support

quozd:
Allows to provide tags overrides using `@swagger_auto_schema` decorator. If overrides not provided, tags will be inferred from the operation url as before.


axnsan12:
Fair enough, thanks!

pptime:
this is great, i just did a workaround by myself

PR #264, opened 2018-12-04
Add support for typed choices

### LINKED ISSUE #260: Detect ChoiceField type based on choices
### Background
According to DRF documentation and source code, `ChoiceField` class supports different values types.
`ChoiceFieldInspector` considers ChoiceField to be of string type in all cases (except ModelSerializer case).

### Goal
Detect field type based on provided choices types.
When all choices are integers, set swagger type to "integer", otherwise use "string".

### Open questions

1. Are there any other types which can be automatically inferred?

mofr:
It resolves #260


mofr:
@axnsan12 Hi, please review my changes and tell if I need to improve it. Thanks.

axnsan12:
Hey, I looked over it, seems good. Also great that you added a test for it, thanks!

PR #268, opened 2018-12-11
Bug fix: return rendered content from _UIRenderer.render instead of HttpResponse

mofr:
According to DRF BaseRenderer documentation BaseRenderer.render should return response content:

> The method should return a bytestring, which will be used as the body of the HTTP response (https://www.django-rest-framework.org/api-guide/renderers/#custom-renderers)

It leads to a very serious problem of double request_finished signal emitting. It happens when drf.Response interprets the HttpResponse returned by render as a file and calls close on it.


mofr:
I can add a regression test for the case if you wish, but it looks quite usual for python language. (So crazy)

axnsan12:
Yeah... good spotting, thanks!

PR #272, opened 2018-12-13
Add support for custom and collection type hint classes

mofr:
### Goal
Support custom user types and more standard types in type hinting.
It's especially useful when DRF JSON encoder is extended with custom types and those types are used as a return type on SerializerMethodField. Also it will work for types like `List[int]`.

#### Collection type hint
```
class SomeSerializer(serializers.Serializer):
    some_field = serializers.SerializerMethodField()
    
    # no need to use a decorator here, it will be an array of integers
    def get_some_field(self, instance) -> List[int]:
        return [1, 2, 3]
```
#### Custom hint class inspector
```
def inspect_enum(enum):
    choices = [entry.value for entry in enum]
    return {
        'type': openapi.TYPE_STRING,
        'enum': choices,
    }

hinting_type_info.append((Enum, inspect_enum))
```
## P.S.
I'm open for any of your suggestions in this direction, so please, tell what I could do better.


mofr:
@axnsan12  I have added several more functions to `inspectors/field.py` file. Would you mind if I go and convert into a package with files like `choice_field_inspector.py`, `serializer_method_field_inspector.py`, `type_hint_class_inspector.py` etc? I will do it only if you like the idea.

src/drf_yasg/inspectors/field.py
```diff
@@ -455,18 +455,89 @@ def decimal_return_type():
...
     (datetime.datetime, (openapi.TYPE_STRING, openapi.FORMAT_DATETIME)),
     (datetime.date, (openapi.TYPE_STRING, openapi.FORMAT_DATE)),
-    # TODO - support typing.List etc
 ]
 
-hinting_type_info = raw_type_info
+
+def get_origin_type(hint_class):
+    origin_type = hint_class.__origin__ if hasattr(hint_class, '__origin__') else None
+    return origin_type or hint_class
```
axnsan12: This can be replaced by `getattr(hint_class, '__origin__', hint_class)`

src/drf_yasg/inspectors/field.py
```diff
@@ -455,18 +455,89 @@ def decimal_return_type():
...
+
+def inspect_union_hint_class(hint_class):
+    if not typing:
+        return NotHandled
+
+    if get_origin_type(hint_class) == typing.Union:
+        child_type = hint_class.__args__[0]
+        return get_basic_type_info_from_hint(child_type)
+
+    return NotHandled
```
axnsan12: This is ugly, we can't just assume by default that an Union is represented by its first alternative. Since OpenAPI 2.0 doesn't provide a good way to express unions (i.e. `oneOf`), we should just ignore them - users can decide exactly how to handle them, if they so wish.

src/drf_yasg/inspectors/field.py
```diff
@@ -455,18 +455,89 @@ def decimal_return_type():
...
+        ])
+
+    return NotHandled
+
+
+hint_class_inspectors = [
+    inspect_union_hint_class,
+    inspect_primitive_hint_class,
+    inspect_collection_hint_class,
+]
```
axnsan12: The addition of this array and the whole inspector logic seems to me like it's making stuff more complicated than it needs to be. 
All the logic can be expressed by adding support for callables instead (type, format) tuples to `hinting_type_info` - i.e. when `hint_class` is subclass of `typing.Sequence`, a function will be called to get type info.

src/drf_yasg/inspectors/field.py
```diff
@@ -455,18 +455,89 @@ def decimal_return_type():
...
+
+
+def register_hint_class_inspector(inspector):
+    """Registers a user defined type hint inspector, which converts it to a swagger type.
+    It can be used for a custom JSON serializable types, like enums.
+
+    :param inspector: a callable of one argument of type hint class.
+    It should return a swagger type keywords arguments dict or NotHandled if the type wasn't handled.
+    """
+    hint_class_inspectors.append(inspector)
```
axnsan12: I don't quite like this. I don't think we should start providing an interface for type hint inspection, I don't see it as something important enough. 

Users can append to the global array themselves, if they so wish, but do so while acknowledging that it's not an explicitly supported extension point.

tests/test_get_basic_type_info_from_hint.py
```diff
@@ -0,0 +1,34 @@
...
+        (List[bool], {'type': openapi.TYPE_ARRAY, 'items': openapi.Items(openapi.TYPE_BOOLEAN)}),
+        (Set[int], {'type': openapi.TYPE_ARRAY, 'items': openapi.Items(openapi.TYPE_INTEGER)}),
+        (Union[List[int], type(None)], {'type': openapi.TYPE_ARRAY, 'items': openapi.Items(openapi.TYPE_INTEGER)}),
+        # Following cases are not 100% correct, but it should work somehow and not crash.
+        (Union[int, bool], {'type': openapi.TYPE_INTEGER, 'format': None}),
+        (List, {'type': openapi.TYPE_ARRAY, 'items': openapi.Items(openapi.TYPE_STRING)}),
+    ])
+    def test_get_basic_type_info_from_hint(hint_class, expected_swagger_type_info):
+        type_info = get_basic_type_info_from_hint(hint_class)
+        assert type_info == expected_swagger_type_info
```
axnsan12: Big like for the comprehensive test!

axnsan12:
Hello!

This is a good initiative, thanks for taking the time to implement and test it!

>Would you mind if I go and convert it to a package with files like choice_field_inspector.py, serializer_method_field_inspector.py, type_hint_class_inspector.py etc? I will do it only if you like the idea.

I don't really like this idea, I don't want to give too much attention to the type hinting feature. It's a nice-to-have for simple cases but should not be overdone. The current structure seems good enough to me.

I've added some comments for specific parts of the code.

src/drf_yasg/inspectors/field.py
mofr: Got it. One more reason has left to work with Union is that `Optional[T]` is represented as `Union[T, None]`. So I will rework it support the only case and leave other non-obvious untouched.

axnsan12: Alright, a special case for `Optional` seems good. In that case it would probably make sense to add `x-nullable` to the type info.

axnsan12:
Thank you for the PR! Bagged and tagged.

PR #278, opened 2018-12-19
add permissions import to quickstart

### LINKED ISSUE #277: `permissions` undefined in Quickstart
Hi folks,

I'm trying to follow the quickstart in README.md, but I'm getting an errot that `permissions` is undefined in `urls.py`.  That makes sense, ad I never imported it, but I don't know where it should be imported from, and the README doesn't mention that either.

The line in the example code is
```
   permission_classes=(permissions.AllowAny,),
```

hoylemd:
I logged #277 when I couldn't run the quickstart, but I figured out how to make it work.

Just adding a line to the quickstart section of the readme so the next person doesn't have the same problem :)

closes #277

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -114,6 +114,7 @@ In ``urls.py``:
 .. code:: python
 
    ...
+   from rest_framework import permissions
    from drf_yasg.views import get_schema_view
    from drf_yasg import openapi
 


axnsan12:
Hey, thanks for the issues, I've amended the quickstart code.

hoylemd:
Awesome! Thanks for the quick follow-up. You rock :+1:

PR #306, opened 2019-02-04
make sure all ListModelMixin views considered as list view

DimasInchidi:
so i realize this weird behavior because i have a view that inherit `generics.ListCreateAPIView` and rendered swagger doesn't add pagination to the response sample for that view. Found that its because [the URL end with parameter](https://github.com/axnsan12/drf-yasg/blob/master/src/drf_yasg/utils.py#L234).

TLDR:
`class GroupMemberListAPIView(generics.ListCreateAPIView)` +
`path('connections/group/<int:group_id>/', GroupMemberListAPIView.as_view())` = not considered as list view


axnsan12:
Makes sense, thanks.

It would be great if you would also add a test for this somewhere, to show the intention and prevent future regressions.

DimasInchidi:
sure, then i'll need to add new view and the url. i'll add endpoint that retrieve list of all objects that related to certain user with url `path('user/model/<int:user_id>/', views.ModelList.as_view())` and for the view, should i add it to [snippets](https://github.com/axnsan12/drf-yasg/blob/master/testproj/snippets/views.py) since it will use generics or to [articles](https://github.com/axnsan12/drf-yasg/blob/master/testproj/articles/views.py) since it will need to have `pagination_class`?

axnsan12:
You can add it anywhere you want, there's no real structure to the test project view, most of them don't even work.

axnsan12:
Looks OK, thanks!

DimasInchidi:
@axnsan12 is this issue already covered by your [last commit](https://github.com/axnsan12/drf-yasg/commit/298a9745df213bedf78fd8e096287fe951ea33a2)?
if yes, i'll close this PR, or do you want to keep test created in this PR?

axnsan12:
He

axnsan12:
The last commit was targeted at #330. I'll be meeting this when I get the chance to rebase it on the latest master (or if you can do that).

Sorry for being so slow about this.

axnsan12:
I meant #331

DimasInchidi:
trigger too many tests, please re-run it. thanks

Issue #307, opened 2019-02-06
Parameter 'items' property interferes with OrderedDict 'items' method


PaulWay:
When trying to define an array parameter, the items definition is stored in an object property named `items`:
```
class Parameter(SwaggerDict):
    def __init__(self, name, in_, description=None, required=None, schema=None,
                 type=None, format=None, enum=None, pattern=None, items=None, default=None, **extra):
        ...
        self.items_ = items
```
However, `items` is the name of a method used by OrderedDicts and other dictionaries to iterate across them.

PaulWay:
Created https://github.com/axnsan12/drf-yasg/pull/308 as a (WIP) fix for this

PR #308, opened 2019-02-06
Store items substructure so as to not conflict with items method

PaulWay:
The items definition in `Items`, `Parameter` and `Schema` needs to be stored in a field that does not conflict with the standard dictionary `items()` method.


axnsan12:
Well, this is quite bad... Thanks for reporting this, I totally missed it!

Changing it like this might completely break some existing code, but at the same time we can't keep hiding the items method. And I can't think of a way to cleanly print a deprecation warning.

I'm going to merge this in and hope not too many people shout at me.

Closes #307, I guess.

Issue #156, opened 2018-07-02
Incorrectly renders CamelCase in all views


glujan:
In my app I'm using [djangorestframework-camel-case](https://github.com/vbabiy/djangorestframework-camel-case) only for some subset of views by setting their parser and renderer classes to `CamelCaseJSONParser` and `CamelCaseJSONRenderer` respectively. My default parser and renderer classes are set to `JSONParser` and `JSONRenderer` in my project settings.

It looks like using camelCase for just one view is causing `drf-yasg` to use camelCase in all views.

axnsan12:
As you can see form the [demo project](https://drf-yasg-demo.herokuapp.com/cached/redoc/#tag/snippets), this is not generally the case. Only views with the parser and renderer you mentioned are converted. Thus it must be something more particular to your project, so please provide some details.

You should note however that due to the nature of OpenAPI and the current implementation of ``drf-yasg``, if you use the same serializer in multiple views, where only some of them are ``camelCase``d, the Swagger model associated with that serializer will be ``camelCase``d everywhere. See also the note [here](https://drf-yasg.readthedocs.io/en/stable/custom_spec.html#inspector-classes). You can work around this by not using named Swagger models ([setting ``ref_name`` to ``None`` on the serializer](https://drf-yasg.readthedocs.io/en/stable/custom_spec.html?highlight=ref_name#serializer-meta-nested-class)).

glujan:
Thank you for detailed response.

I'm indeed reusing some serializers and your note about `ref_name` was what I was missing - setting it to `None` or to string to not clash with original serializer solves the issue.

glujan:
Note: this also happens if I'm not reusing serializers but there is a name clash in across different Django applications. Setting different `ref_name` helps with those issues as well.

Maybe `drf-yasg` should show a warning about that.

axnsan12:
> Maybe drf-yasg should show a warning about that or at least allow to prefix model name with app name.

Yeah, both of those seem reasonable.

PR #313, opened 2019-02-14
Use unambiguous ref names

kammala:
serializer type name may be ambiguous, e.g. in different applications of the same project one may have different `Account` with the same serializer class name but with different fields, etc.
In current implementation they will have the same `$ref` name in API spec which will lead to improper API representation.
Adding module name to ref name should solve this problem.

### DIFF
--- a/src/drf_yasg/utils.py
+++ b/src/drf_yasg/utils.py
@@ -415,7 +415,7 @@ def get_serializer_ref_name(serializer):
         logger.debug("Forcing inline output for ModelSerializer named 'NestedSerializer':\n" + str(serializer))
         ref_name = None
     else:
-        ref_name = serializer_name
+        ref_name = '{module}.{name}'.format(module=serializer.__module__, name=serializer_name)
         if ref_name.endswith('Serializer'):
             ref_name = ref_name[:-len('Serializer')]
     return ref_name
--- a/testproj/articles/serializers.py
+++ b/testproj/articles/serializers.py
@@ -4,6 +4,10 @@ from rest_framework import serializers
 from articles.models import Article, ArticleGroup
 
 
+class OtherStuffSerializer(serializers.Serializer):
+    bar = serializers.CharField()
+
+
 class ArticleSerializer(serializers.ModelSerializer):
     references = serializers.DictField(
         help_text=_("this is a really bad example"),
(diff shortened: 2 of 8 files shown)


rsichny:
Does this mean that for all existing projects all the model names will change? If so i'd say it's not acceptable since it will break *every* client app that uses existing apis.

kammala:
Hello, @rsichny , thanks for your feedback!
Actually, this is a very good question and I don't know complete answer to it.

Yes, it will change names for all models in existing projects, but these names are not part of API itself, so I don't think it should not break anything.

Once generated with `swagger-codegen` client will not revise `swagger.json`, so these changes should not be a problem in this case.

rsichny:
But this means that whenever someone generates a new client using codegen - they'll need to change ALL the code which uses the generated client.

Example 1: in our case QAs generate the new client using codegen for api schemas on each release and then they run test suite (thouthands of test cases) using the newly generated client. Global change of the names which are being output by codegen will produce enormous amount of work (they'll need to change each of the references to the old names).

Example 2: our partners use swagger definitions to generate the swagger client and use that client in their code. As soon as we decide to add a new field to api they'll need to re-generate the swagger client and all the integration will break.

My suggestion is to leave the old behavior by default, and either make this new behavior configurable, or only add prefixes for the cases when the schema is added to the list of definitions and the same definition already exist.

kammala:
However in case of dynamic clients like [`bravado`](https://github.com/Yelp/bravado/blob/f8f98c00c3422f977ae44b4ef7e37350b05f6ea5/bravado/client.py#L138) it will be definitely a problem(don't know if it will be a problem for [`swagger-js`](https://github.com/swagger-api/swagger-js), doesn't seem so)

rsichny:
Also, since all the name conflicts can be resolved by using ref_name in serializer i don't think any change is necessary at all.

https://drf-yasg.readthedocs.io/en/stable/custom_spec.html#serializer-meta-nested-class

kammala:
Yes, I was also thinking about adding some settings to switch this behaviour on keeping old one as a default.

However I still believe that it would be nice to have such unambiguous behavior by default, maybe in the next major release. Right now this behavior is very awkward: you will get improper API documentation in case on names clash without any warning and it is really hard to debug.

I will leave this PR open for now to gather more opinions and have a look if it is possible to make this problem more clear(raise a warning when app finds ambiguous name, e.g.)

rsichny:
@kammala how about adding a warning into https://github.com/axnsan12/drf-yasg/blob/master/src/drf_yasg/openapi.py#L540 when the ref names overlap?

axnsan12:
Others may use the swagger spec for live API clients or form generators, so changing stuff suddenly like this is still kinda awkward.

Plus, the problem this solves is not that common, I think. This solution might work behind a setting. Maybe a better one would be to always raise an exception when encountering multiple serializers with the same name, forcing explicit disambiguation by the user.

axnsan12:
The warning should already be there:

https://github.com/axnsan12/drf-yasg/blob/583e404ed826416bade3505d1c77dba36b98ba9b/src/drf_yasg/inspectors/field.py#L130

rsichny:
@axnsan12 maybe just change to `warnings.warn` to make it more visible? (people often turn off logging for 3rd party libs)

axnsan12:
I think I'll just change it to an exception :smile:

kammala:
However, this `$ref` is still not required in OpenAPI — we can either create definitions for schemas or put these definitions in place they are needed. I think, ideally this part should not be treated as a part of API cause it does not affect actual client-server protocol.

this warning doesn't work for some reason :(

kammala:
`actual_serializer` is always `None` because of https://github.com/axnsan12/drf-yasg/blob/master/src/drf_yasg/openapi.py#L109

so this check will never work :(

kammala:
looks like #156 wasn't fixed

PR #321, opened 2019-02-25
Support RecursiveField with 'many' option

rsichny:
suddenly, we needed this for the case when recursive field was actually referencing not the parent serializer, but the one defined later in the same module


axnsan12:
Seems good, thanks!

PR #334, opened 2019-03-20
add operation_keys to get_summary_and_description method

Amoki:
I use `operation_keys` to override the `get_operation_id` method and be able to configure my operation ids with less overhead than method_decoration + swagger_decorator + operation_id.
eg: 
```
class UserViewSet(ModelViewSet):
    queryset = User.objects.all()
    operations = {
        "list": "getUsers",
        "read": "getUser",
        "create": "createUser",
        "delete": "deleteUser",
    }
```
I'd like to do something similar with the description and summary for each path but the method `get_summary_and_description` don't have access to the operation_keys.

### DIFF
--- a/src/drf_yasg/inspectors/view.py
+++ b/src/drf_yasg/inspectors/view.py
@@ -33,7 +33,7 @@ class SwaggerAutoSchema(ViewInspector):
         parameters = self.add_manual_parameters(parameters)
 
         operation_id = self.get_operation_id(operation_keys)
-        summary, description = self.get_summary_and_description()
+        summary, description = self.get_summary_and_description(operation_keys)
         security = self.get_security()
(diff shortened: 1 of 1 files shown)


axnsan12:
Can't you use `self.view.action`?

Amoki:
Unfortunately, retrieving the action is more complex than that: https://github.com/encode/django-rest-framework/blob/master/rest_framework/schemas/generators.py#L412
Operation keys have also more information

axnsan12:
Fair enough. In that case, I would rather move the `operation_keys` parameter to the `SwaggerAutoSchema` constructor and set it as an instance attribute, to make it available to any method which might need it without changing all signatures.

Any change to method signatures must also take backward compatibility into account, both for callers and for overriders.

PR #337, opened 2019-03-25
Return true in is_list_view for ListModelMixin

parander:
Examples for ListAPIView's sometimes don't get the response shown as an array ([]) in the documentation due to is_list_view guessing that the view is not returning a list.

### DIFF
--- a/src/drf_yasg/utils.py
+++ b/src/drf_yasg/utils.py
@@ -6,7 +6,7 @@ from collections import OrderedDict
 from django.db import models
 from django.utils.encoding import force_text
 from rest_framework import serializers, status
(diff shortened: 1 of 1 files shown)


axnsan12:
Please provide an example of this happening.

axnsan12:
Duplicate of #306

parander:
Every ListAPIView that ends with some parameter will:

`    # if the last component in the path is parameterized it's probably not a list view
    path_components = path.strip('/').split('/')
    if path_components and '{' in path_components[-1]:
        return False
`

e.g.
`path('mylist/<int:id>',
        SomeListAPIView.as_view(),
        name='mylist',
)`

PR #357, opened 2019-05-01
'decimal' format is invalid in combination with TYPE_NUMBER

ayushin:
As per openapi specification type 'number' does not allow 'decimal' format and this breaks openapi-generator. This workaround sets format to None if the type is not 'string' so that it is recognised as an arbitrary number by the openapi-generator, but still allows using 'decimal' format when Decimal is coerced to string. Another solution would be setting FORMAT_DECIMAL to None in openapi.py but I think it would be less flexible and might break existing implementations.


axnsan12:
As per the [OpenAPI specification](
#include "falx_constants.h"), the `format` property is a free-form string whose semantics can be chosen arbitrarily by implementations. The spec then gives a non-exhaustive list of reasonable formats to support, but it is completely correct to generate other values too.

PR #363, opened 2019-05-07
Fix unicode attribute error if typing backport installed on python2.7

elliott-omosheye:
Python2.7 with Typing backport module installed throws `AttributeError: type object 'unicode' has no attribute '__args__'` this is because this commit https://github.com/axnsan12/drf-yasg/commit/7bb4700003561b7bb6846d69866e69452830b916 calls `json.loads(json.dumps(value))` which returns unicode strings in Python2.7 and the `hinting_type_info` does not have unicode in the list (assuming because unicode is no longer a class in Python3) but if the typing module is installed it will think that the unicode type class is a collection and will call `inspect_collection_hint_class` in field.py.

Also I had to change the testproject slightly in order to get the tests to run with the typing backport installed as the typing style is not supported on python2

Not sure what the correct way of checking Python version is but I think there is enough information here to get a better solution if this is not the correct style for this project.

### DIFF
--- a/src/drf_yasg/inspectors/field.py
+++ b/src/drf_yasg/inspectors/field.py
@@ -2,6 +2,7 @@ import datetime
 import inspect
 import logging
 import operator
+import sys
 import uuid
 from collections import OrderedDict
 from decimal import Decimal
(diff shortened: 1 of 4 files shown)


axnsan12:
Seems OK, thank you!

PR #364, opened 2019-05-11
update reference.yaml based on pygments lexers & styles changes

terencehonles:
Snippet model in testproj/snippets/models.py has field choices which depend on ``pygments.lexers`` and ``pygments.styles``


axnsan12:
Thanks!

PR #387, opened 2019-06-17
force encoding

FilArt:
docs:
изображение

results "python manage.py generate_swagger file.yml" below:

1) before changes:
изображение

2) after changes:
изображение


axnsan12:
I don't think this is a good idea. Ensuring escaped ascii output in JSON and YAML helps compatibility.

You would only care abut this if you 

- use broken tools
- want to look directly read the yaml file as documentation
- care about file size

For all of those cases the burden is on the user to properly convert their data into a more appropriate encoding, a task which is made easier by using non-ambiguous ASCII output.

PR #399, opened 2019-07-03
Get Parameter and Response descriptions from serializer docs if available

PaulWay:
This provides descriptions in all `openapi.Parameter` and `openapi.Response` objects that are based on `serializer` objects.  The description is based on the serializer's docstring.  If the serializer is not provided, or it has no docs, the empty string is used (which is the previous default operation).

Signed-off-by: Paul Wayper <paulway@redhat.com>


JoelLefkowitz:
Shadowed by #786

PR #400, opened 2019-07-04
handle enum type for nested ChoiceFields

etene:
Hello,

On a project I'm working on, I'm using Django `IntegerField`s, with a `choices` parameter, nested in postgres `ArrayFields`. This translates to a `ListField` with a child `ChoiceField` in the `ModelSerializer`.

The issue I'm having is that `ChoiceFieldInspector` fails to properly infer the resulting enum type, even though it could because it can guess it correctly for `IntegerField`s:

- First, it tries to get the corresponding model field, but since the nested `ChoiceField` has no `source`, it fails. I fixed it by using `parent.source`.
- Then, even if it manages to get the corresponding model field, it's an `ArrayField` so `get_basic_type_info` doesn't return anything. I fixed it by using the `base_field` if present.

This PR adds some code that fixes the issue & a test. Feel free to tell me if I solved the problem the wrong way; tests seem to pass so I don't think I broke anything.

### DIFF
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -5,5 +5,7 @@ pytest-cov>=2.6.0
 pytest-xdist>=1.25.0
 pytest-django>=3.4.4
 datadiff==2.0.0
+psycopg2-binary==2.8.3
+django-fake-model==0.1.4
 
 -r testproj.txt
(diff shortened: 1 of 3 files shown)


axnsan12:
Seems good, thank you!

PR #408, opened 2019-07-13
Fix imports for upcoming DRF version

tfranzel:
Came across this while working on a PR for DRF. Upstream changes to DRF will break drf-yasg with the release of [3.10](https://github.com/encode/django-rest-framework/tree/version-3.10). DRF is dropping Py2 support and did some refactorings.

Tests worked fine with DRF branch `version-3.10` installed in the virtualenv

[breaking DRF PR](https://github.com/encode/django-rest-framework/pull/6532)

Sidenote: I also had to create a PR for the dependency [djangorestframework-camel-case](https://github.com/vbabiy/djangorestframework-camel-case/pull/58) as it will also break on 3.10.

### DIFF
--- a/src/drf_yasg/generators.py
+++ b/src/drf_yasg/generators.py
@@ -4,14 +4,22 @@ import re
 from collections import OrderedDict, defaultdict
 
 import uritemplate
+import rest_framework
 from coreapi.compat import urlparse
 from rest_framework import versioning
 from rest_framework.compat import URLPattern, URLResolver, get_original_route
-from rest_framework.schemas import SchemaGenerator
 from rest_framework.schemas.generators import EndpointEnumerator as _EndpointEnumerator
 from rest_framework.schemas.generators import endpoint_ordering, get_pk_name
-from rest_framework.schemas.utils import get_pk_description
+
 from rest_framework.settings import api_settings
+
+from packaging.version import Version
+if Version(rest_framework.__version__) < Version('3.10'):
+    from rest_framework.schemas.generators import SchemaGenerator
+    from rest_framework.schemas.inspectors import get_pk_description
+else:
+    from rest_framework.schemas import SchemaGenerator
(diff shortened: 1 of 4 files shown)


axnsan12:
Thanks for the initative, but your changes seem to break compatibility with older versions: https://travis-ci.org/axnsan12/drf-yasg/builds/558380899 (not sure why the PR status was not updated by Travis)

tfranzel:
Sure. I will have a look in an hour. DRF just released 3.10 a couple of hours ago. Yasg is broken as of now. A coworker just ran into exactly this issue 5 min ago :)

tfranzel:
ok took longer than anticipated but i got it done. there are multiple things: 

- to keep compatibility with old versions I needed to add an version check for >3.10 and below.
- to make the tests work with 3.10 (and below) i had to add my PR of [djangorestframework-camel-case](https://github.com/vbabiy/djangorestframework-camel-case/pull/58) to the testproj requirements until its gets merged. As of now the lib is broken.
- added tox target for `drf310`
- `djmaster` is only failing because `corsheaders` also uses `from django.utils import six` which got removed from master. honestly, didn't want to open up a third PR.

axnsan12:
Awesome, thanks!

src/drf_yasg/generators.py
```diff
@@ -4,15 +4,23 @@
...
 from rest_framework import versioning
 from rest_framework.compat import URLPattern, URLResolver, get_original_route
-from rest_framework.schemas import SchemaGenerator
 from rest_framework.schemas.generators import EndpointEnumerator as _EndpointEnumerator
 from rest_framework.schemas.generators import endpoint_ordering, get_pk_name
-from rest_framework.schemas.utils import get_pk_description
+
 from rest_framework.settings import api_settings
 
+from packaging.version import Version
```
sycdan: This broke our build with `ModuleNotFoundError: No module named 'packaging'`

PR #409, opened 2019-07-15
fixed endpoint listing when using URLPathVersioning

trolley813:
When using URLs which are valid only for a specific set of API versions, like this:
```python
# this is present only in v1
url(r'^api/(?P<version>v1)/one/', include('app1.urls')),
# and this is only in v2
url(r'^api/(?P<version>v2)/two/', include('app2.urls')),
```
the Swagger/Redoc generator seems to generate docs for all endpoints and all versions (i.e. in the example above, `/api/v1/two` and `/api/v2/one` endpoints get included despite the fact they are in fact nonexistent), when using `URLPathVersioning` as default versioning class (since the version regex part is replaced by `{version}`, regardless of the actual restriction on version in the URL, so it gets included).

This PR fixes the bug (if it's a bug at all), by parsing actual version regex and matching the request version with it.


src/drf_yasg/generators.py
```diff
@@ -89,8 +102,12 @@ def get_api_endpoints(self, patterns=None, prefix='', app_name=None, namespace=N
                 try:
                     path = self.get_path_from_regex(path_regex)
                     callback = pattern.callback
+                    versioning_class = getattr(callback.cls, 'versioning_class', None)
+                    version = getattr(self.request, 'version', None)
+                    version_param = version and getattr(versioning_class, 'version_param', 'version')
+                    version_regex = self.get_version_regex_from_path_regex(path_regex, version_param)
```
axnsan12: Why are this lines here and not in `should_include_endpoint`?

src/drf_yasg/generators.py
```diff
@@ -34,15 +34,28 @@ def get_path_from_regex(self, path_regex):
             logger.warning("url pattern does not end in $ ('%s') - unexpected things might happen", path_regex)
         return self.unescape_path(super(EndpointEnumerator, self).get_path_from_regex(path_regex))
 
-    def should_include_endpoint(self, path, callback, app_name='', namespace='', url_name=None):
+    def get_version_regex_from_path_regex(self, path_regex, version_param):
```
axnsan12: Will this function also work with Django 2.0 `path`? i.e. `path('api/<string:version>/one', include(...))`

src/drf_yasg/generators.py
```diff
@@ -34,15 +34,28 @@ def get_path_from_regex(self, path_regex):
...
+        if path_regex.endswith(')'):
+            logger.warning("url pattern does not end in $ ('%s') - unexpected things might happen", path_regex)
+        version_param_regex = re.compile(rf"\(\?P<{version_param}>(?P<value>.+?)\)")
```
axnsan12: You can't use an f-string here since we still support Python 2.7 and 3.5.

src/drf_yasg/generators.py
hyst329: Thank you for pointing! Fixed.

src/drf_yasg/generators.py
hyst329: I'm unsure if `URLPathVersioning` works with 2.0 paths at all.

src/drf_yasg/generators.py
hyst329: Fixed, rolling back to the good old `%` operator.

src/drf_yasg/generators.py
axnsan12: Of course it does, and it would be considered broken if it didn't. It is evident from the Django and DRF docs that `path`, possibly with [custom converters](https://docs.djangoproject.com/en/2.2/topics/http/urls/#registering-custom-path-converters), is the preferred path templating mechanism going forward. 

`re_path`/`url` should only be used for compatibility, or when absolutely necessary.

hyst329: Updated for standard path converters - `string`, `int` and `slug` (well, it's weird to use other ones like `uuid` or `path` for versioning, and custom ones are not supported here (in `drf-yasg`) at all, as I see).

axnsan12:
Tests seem to be failing now.

PR #417, opened 2019-07-19
inspectors: add support for JSONField

etene:
Hello,

I needed `JSONField`s to be properly represented as `object`s in my generated api spec, but there was no `JSONFieldInspector`, so I added one, plus a test.

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -85,6 +85,7 @@ to this list.
 :class:`'drf_yasg.inspectors.ChoiceFieldInspector' <.inspectors.ChoiceFieldInspector>`, |br| \
 :class:`'drf_yasg.inspectors.FileFieldInspector' <.inspectors.FileFieldInspector>`, |br| \
 :class:`'drf_yasg.inspectors.DictFieldInspector' <.inspectors.DictFieldInspector>`, |br| \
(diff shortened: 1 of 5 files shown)


axnsan12:
JSONField is not necessarily an object - it can also be a list or a primitive.

axnsan12:
However, I guess that object is the most common case so it would be good to make that the default instead of string.

etene:
Thanks, I figured the same; the type can still be overridden if necessary but in most cases it will probably be `object`.

I could't think of a way to guess the expected value anyway, except maybe guessing it from the field's default value if any. But that would make it inconsistent depending on whether there is a default value or not, which is better avoided.

johnthagen:
>  the type can still be overridden if necessary

@etene Can you elaborate on how you would do this for a single `JSONField`? I've been trying to use the [`Meta` field option](https://drf-yasg.readthedocs.io/en/stable/custom_spec.html#serializer-meta-nested-class) but haven't been able to figure out how to do that from the docs.

etene:
- you could use a custom `FieldInspector` that returns your wanted schema (and give it priority over the default `JSONFieldInspector`)
- or if it's for a single instance, leverage serializer introspection and use a custom serializer (as mentioned in #478)

Issue #426, opened 2019-07-29
Optional type hint schema generation support broken


johnthagen:
1.15.1 broke support for `Optional` type hinting schema generation.

- Support for `Optional` was added in #272 
- I believe that support was broken in #363 or 69b628a7afdbc9de5f8872a34669a4b1a827155b (part of 1.15.1).

Python versions on which the problem as been reproduced:
- CPython 3.7.3
- CPython 3.6.8
- PyPy 7.1.1 (3.6 compatible)

To reproduce, create a `Serializer` like:

```py
from typing import Optional

class MySerializer(Serializer):
    x = SerializerMethodField()

    def get_x(self, instance) -> Optional[int]:
        pass
```

In 1.15.0, this is properly detected in the generated schema as an integer.
In 1.16.0 (I couldn't get 1.15.1 to run properly), it is incorrectly detected as a string.

johnthagen:
cc'ing @xavfernandez in case he has insights from his PR.

etene:
I think you're right, it looks like it was broken in 69b628a by removing the `if not inspect.isclass(hint_class) and hasattr(hint_class, '__args__'):` check in `SerializerMethodFieldInspector`. I have a PR coming up.

johnthagen:
#428 has been merged. Closing.

PR #428, opened 2019-08-01
fix Optional typing hint for SerializerMethodField

etene:
Hello,

As reported in #426, commit 69b628a broke type hinting for `SerializerMethodField` return values.
Here is a fix`that reverts two lines that I believe should not have been removed, and adds a test to ensure it won't break again.

I only tested with 3.7 on my machine, I hope it doesn't break anything for other Python versions.


etene:
The test breaks Python 2.7 compatibility, I'll try to fix it.

pahaz:
is it going to be merged? I have the same issue

axnsan12:
Seems good, thanks!

Issue #423, opened 2019-07-26
Issue with ruamel.yaml


nicholasgcoles:
Getting an issue when trying to build my dockerfile

The error is as follows
```
 from drf_yasg.views import get_schema_view
  File "/usr/local/lib/python3.7/site-packages/drf_yasg/views.py", line 14, in <module>
    from .renderers import (
  File "/usr/local/lib/python3.7/site-packages/drf_yasg/renderers.py", line 11, in <module>
    from .codecs import VALIDATORS, OpenAPICodecJson, OpenAPICodecYaml
  File "/usr/local/lib/python3.7/site-packages/drf_yasg/codecs.py", line 133, in <module>
    class SaneYamlDumper(yaml.SafeDumper):
AttributeError: module 'ruamel.yaml' has no attribute 'SafeDumper'
```

I think this could be because the `ruamel.yaml` library updated yesterday to `0.16.0`

I hardcoded the `ruamel.yaml` dependency to be `ruamel.yaml==0.15.X` and then my docker container proceeded to build correctly.

I think a fix for now could be updating your base.txt to:
`ruamel.yaml>=0.15.34,<0.16.0`

Almaz97:
Yeah, the reason was as supposed. Thanks a lot!

cjmaynar:
I'm continuing to see this issue after adding that requrement to my environment. I currently have
```
ruamel.yaml==0.15.100
ruamel.yaml.clib==0.1.0
```

Is there any other recommended temporary fix for this?

dadaday:
I also stumbled upon this error.
Removing the python image from docker and rebuilding the container solved the problem.
I've checked the installed packages inside django container and it is working with `ruamel.yaml==0.16.0`

axnsan12:
Seems to have been fixed in later versions of `ruamel.yaml` (0.16.5 works)

PR #430, opened 2019-08-01
Avoid moving up to ruamel.yaml 0.16.x

ashokdelphia:
Following the usual semver rules, pre 1.0 packages may have breaking changes between minor versions, so it makes sense to pin to >=0.15.34 and <0.16.0, which is most easily expressed with the "compatible release" operator, ~=.

See #423.


ashokdelphia:
It looks like this is no longer breaking with 0.16.1.

Issue #386, opened 2019-06-17
Wrong schema for FileField in model with default parser classes


maun:
When using a FileField in the model, a ModelSerializer, and a ModelViewSet DRF generates a endpoint where the file is uploaded using the MIME type multipart/form-data.

drf-yasg does not recognize this and instead skips the File field in the request but includes the other parameters in the scheme as JSON body parameters.

When setting the parser classes `parser_classes = [FormParser, MultiPartParser]` on the viewset the correct scheme is generated using form parameters. Unfortunately this is not an option because the update endpoint of the ModelViewSet requires the JSONParser.
Also adding the parameter manually as a form parameter does not work, this error is thrown: https://github.com/axnsan12/drf-yasg/blob/e9f27442fce1ce6df7aa33117e850b63a86a4051/src/drf_yasg/inspectors/view.py#L166

daviddavis:
I'm dealing with the same problem. However, it looks like the accepted mimetypes for an endpoint that accepts a file upload is json **and** multipart/form-data:

```
(Pdb) request.parsers
[<rest_framework.parsers.JSONParser object at 0x7f2eb3352b00>, <rest_framework.parsers.FormParser object at 0x7f2eb3352b38>, <rest_framework.parsers.MultiPartParser object at 0x7f2eb3352b70>]
```

I assume that somehow drf-yasg is picking the JSONParser and then excluding the FileField because it can't be submitted via json.

You can customize the parser_classes with the `@action` decorator but I don't think that it's available for predefined routes like create, update, etc. At least, I get the error: `Cannot use the @action decorator on the following methods, as they are existing routes: create` when I try to do so.

Perhaps drf-yasg could add an option to @swagger_auto_schema where you could manually set the `consumes` media type?

daviddavis:
This patch worked for me although I am sure there's a better way to do this:

```diff
diff --git a/src/drf_yasg/inspectors/view.py b/src/drf_yasg/inspectors/view.py
index 334c274..c7d8c06 100644
--- a/src/drf_yasg/inspectors/view.py
+++ b/src/drf_yasg/inspectors/view.py
@@ -29,6 +29,12 @@ class SwaggerAutoSchema(ViewInspector):
         consumes = self.get_consumes()
         produces = self.get_produces()
 
+        multipart = ['multipart/form-data', 'application/x-www-form-urlencoded']
+        if 'file' in [param['type'] for param in self.get_request_body_parameters(multipart)]:
[...]

PR #436, opened 2019-08-07
Support multipart media type by default for filefield/imagefield

ychab:
See issue https://github.com/axnsan12/drf-yasg/issues/386

With that fix, it means that we may always use Multipart media type instead of JSON parser... So not really sure if it is a good idea or not :/

A better alternative would be to have to select the content type in the UI of Swagger or Redoc instead....

### DIFF
--- a/src/drf_yasg/utils.py
+++ b/src/drf_yasg/utils.py
@@ -374,8 +374,12 @@ def get_consumes(parser_classes):
     parser_classes = [pc for pc in parser_classes if not issubclass(pc, FileUploadParser)]
     media_types = [parser.media_type for parser in parser_classes or []]
     non_form_media_types = [encoding for encoding in media_types if not is_form_media_type(encoding)]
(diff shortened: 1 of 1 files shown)


ychab:
Oh, I probably know why this is not possible and why it decide to switch on application/json media type by default if we don't have only form media type : this is because multipart/form-data doesn't allow nested parameter... !!
I just refresh the pull request in order to add some comments about this because this is really tricky and it could help to understand why (if this is the good reason anyway)

axnsan12:
Yeah, your comments pretty much hit it on the head. The problem is both that there is no widely accepted standard for nested values in multipart/form-data, and that OpenAPI 2.0 enforces this separation via Parameters for form fields and Schemas for json body.

OpenAPI 3 is a real big step forward in this regard since it allows all request body types to be specified as a Schema, with a lot of knobs for media types. However building a lib to generate that is probably a nightmare.

PR #444, opened 2019-08-22
custom_spec.rst: Update to indicate swagger_auto_schema import location.

grumbling-tom:
At the moment, the docs reference the `@swagger_auto_schema` decorator without indicating how to import it into an application - this PR adds in that information.


axnsan12:
Thanks!

PR #461, opened 2019-09-26
[WIP] Merge in features from DRF core

kevin-brown:
This is a proof of concept to see if it's even possible. If this ends up being possible, the intention is to layer in smaller pull requests which are targeted instead of a massive one (so, not using this branch).


kevin-brown:
I think at this point I can definitely say that there are some easy wins for this project in terms of reducing code duplication and aligning with how DRF does things. There are definitely some parts where we have forked off so much that it's going to be hard to come back together, but I'm less concerned about those parts for right now.

I have also confirmed two bugs that were fixed here but still exist in DRF that can be corrected.

My plan is to start layering in small pull requests next time I come back to this, likely covering the following first:

* `SwaggerValidationError` can be made to subclass `APIException`, removing the need for `SwaggerExceptionMiddleware`
* `is_callable_method` is a more complex version of `is_simple_callable` that DRF uses
* `EndpointEnumerator` still has overrides for bugs in DRF versions that are no longer supported
* The `format` used by the renderers are non-standard (leading `.` should be removed)

---

@axnsan12 it looks like Travis CI is no longer reporting the commit statuses

marco-silva0000:
I noticed that DRF's web api sets the title of the page based on the View class name, https://www.django-rest-framework.org/topics/documenting-your-api/#setting-the-title

Would this merge sync this functionality?

PR #464, opened 2019-09-30
Apply dedent to descriptions provided in the swagger_auto_schema decorator

### LINKED ISSUE #416: Descriptions should be textwrap.dedent'ed
Any place in the decorators that a description can be provided, the string should be passed through textwrap.dedent.  This way, we can write multi-line strings, keep them indented to look right in the decorator, and still get the correct markdown in the result.

yurihs:
Hi! This should fix #416.

Please let me know if I missed anything.

### DIFF
--- a/src/drf_yasg/utils.py
+++ b/src/drf_yasg/utils.py
@@ -1,6 +1,7 @@
 import inspect
 import logging
 import sys
+import textwrap
 from collections import OrderedDict
 from decimal import Decimal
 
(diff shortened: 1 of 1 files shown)


axnsan12:
It would probably be better to do this at a deeper level, where the values are actually used. Most likely you could just throw decent onto all usages of `force_real_str`.

yurihs:
Yeah, that seems better. I was wary of changing the behavior of `force_real_str`.

kevin-brown:
Looks like tests weren't updated with this, so now it's failing tests on master: https://travis-ci.com/axnsan12/drf-yasg/builds/136509688

Issue #439, opened 2019-08-12
Write out multiline strings as yaml block scalars?


nedbat:
Is it possible to write out multi-line descriptions (and other strings) as YAML block scalars? The resulting file would be more readable.

For example, instead of:
```
paths:
  /api/one:
    post:
      operationId: api_one
      summary: Inserts a batch of completions.
      description: "REST Endpoint Format:\n{\n  \"username\": \"username\",\n  \"\
        course_key\": \"course-key\",\n  \"blocks\": {\n    \"block_key1\": 0.0,\n\
        \    \"block_key2\": 1.0,\n    \"block_key3\": 1.0,\n  }\n}\n\n**Returns**\n\
        \nA Response object, with an appropriate status code.\n\nIf successful, status\
        \ code is 200.\n{\n   \"detail\" : _(\"ok\")\n}\n\nOtherwise, a 400 or 404\
        \ may be returned, and the \"detail\" content will explain the error."
      parameters: []
      responses:
        '201':
          description: ''
      tags:
        - api
    parameters: []
```
write out the description like this:
```
paths:
  /api/one:
    post:
      operationId: api_one
      summary: Inserts a batch of completions.
      description: |
        REST Endpoint Format:
        {
          "username": "username",
          "course_key": "course-key",
          "blocks": {
            "block_key1": 0.0,
            "block_key2": 1.0,
            "block_key3": 1.0,
          }
        }

        **Returns**

        A Response object, with an appropriate status code.

        If successful, status code is 200.
        {
           "detail" : _("ok")
        }

        Otherwise, a 400 or 404 may be returned, and the "detail" content will explain the error.
      parameters: []
      responses:
        '201':
          description: ''
      tags:
        - api
    parameters: []
```

The data is the same other than a trailing newline on the description.

axnsan12:
This would be nice to have, however I don't know how to tickle the yaml code to produce this. It would be great if you could propose the necessary changes to `codecs.SaneYamlDumper`.

nedbat:
Looks like this Stack Overflow question has answers about how: https://stackoverflow.com/a/33300001/14343
```
import yaml

def str_presenter(dumper, data):
  if len(data.splitlines()) > 1:  # check for multiline string
    return dumper.represent_scalar('tag:yaml.org,2002:str', data, style='|')
  return dumper.represent_scalar('tag:yaml.org,2002:str', data)

yaml.add_representer(str, str_presenter)
```
I might get a chance to try it out locally.

[...]

PR #466, opened 2019-10-01
Write multi-line strings in block style. #439

nedbat:
(I don't have new tests, but the tests pass.)

### DIFF
--- a/src/drf_yasg/codecs.py
+++ b/src/drf_yasg/codecs.py
@@ -1,5 +1,3 @@
-from six import raise_from
-
 import copy
 import json
 import logging
@@ -7,6 +5,8 @@ from collections import OrderedDict
 
 from coreapi.compat import force_bytes
 from ruamel import yaml
+
+from six import binary_type, raise_from, text_type
 
 from . import openapi
 from .errors import SwaggerValidationError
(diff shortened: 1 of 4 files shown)


src/drf_yasg/codecs.py
```diff
@@ -176,7 +176,14 @@ def represent_odict(self, mapping, flow_style=None):  # pragma: no cover
                 node.flow_style = best_style
         return node
 
+    def represent_text(self, text):
+        if len(text.splitlines()) > 1:
```
axnsan12: It would likely be more efficient to use `'\n' in text or '\r' in text` here. Serializing to YAML is already very slow so let's not make it worse than needed.

src/drf_yasg/codecs.py
```diff
@@ -176,7 +176,14 @@ def represent_odict(self, mapping, flow_style=None):  # pragma: no cover
...
+            return self.represent_scalar('tag:yaml.org,2002:str', text, style='|')
+        return self.represent_scalar('tag:yaml.org,2002:str', text)
 
+
+SaneYamlDumper.add_representer(str, SaneYamlDumper.represent_text)
+SaneYamlDumper.add_representer(unicode, SaneYamlDumper.represent_text)
```
axnsan12: Will this work on python 3?

axnsan12: Answer is no: https://travis-ci.org/axnsan12/drf-yasg/jobs/592217419

nedbat:
Yeah, sorry, I got too excited and went too fast :)   This should now be good for Python 3, and includes a test to verify the behavior.

src/drf_yasg/codecs.py
nedbat: Done

PR #470, opened 2019-10-04
Subclass SchemaGenerator instead of defining a new one

kevin-brown:
While the SchemaGenerator provided by DRF-YASG is not API-compatible with the one used by DRF, it does use a lot of the methods and even generates a DRF one internally. Given that this required quite a bit of extra code to be written to do the wrapping, it makes sense to instead just subclass it and get those additional benefits.

This should make it easier to migrate from the CoreAPISchemaGenerator, which is what is currently being used, to the DRF-provided OpenAPISchemaGenerator for the internal serialization. That was not done here because there is still a lot of work that needs to be done in order to enable that.


PaulWay (approved):
I think this is a good idea.

Issue #481, opened 2019-10-27
Type is incorrect if the lookup_field is set to be a one-2-one relation


lingster:
Create say a UserProfile class which has a One2One relation to the User class, where the id is an integer. 

    class UserProfile():
        user = models.OneToOneField(User, on_delete=models.CASCADE, primary_key=True)

If you define the view set with `lookup_field='user_id'` then the generated swagger doc defaults the lookup_field as a string instead of integer
    class UserProfileViewSet(viewsets.GenericViewSet):
        queryset = UserProfile.objects.all()
        serializer_class = UserProfileSerializer
        lookup_field = 'user_id'
    
I'm currently working on a PR for this:

PR #482, opened 2019-10-27
For OneToOne lookup field use the target_field to determine type

lingster:
fix for #481


lingster:
@axnsan12 : seems like the tox env djmaster is failing as django 3.1 has the following update:
`The compatibility import of django.core.exceptions.FieldDoesNotExist in django.db.models.fields is removed.`

It seems that there are some external packages that are still doing from `django.db.models.fields import FieldDoesNotExist` which is causing the CI to fail. Is there a work around for this or should we just waiting until other projects have been updated?

PR #483, opened 2019-10-28
Adds support to DEFAULT_JSON_ENCODER

gustav0:
Hello,

I've been using custom JSON encoders for a while with drf-yasg but haven't seen any support for it.

A while ago someone requested support for the django Decimal field and it was implemented. But i think this might be a simple solution if people want to use their own fields without needing to add custom code, just use their own encoder (which they probabbly they already have if they are using drf).

I'm willing to make any changes and validations you guys suggest and update the documentation.


creyD:
Screenshot 2022-07-19 at 09 01 01

Thank you for your PR! You only replaced it in utils, shouldn't the instances in renderers also be replaced? Also the docs will have to be updated for this additional setting.

Issue #485, opened 2019-10-30
Error calling generate_swagger command with custom user model


CarlosMart626:
If the project has defined a custom model with a custom username field this command will return this error.

```
python manage.py generate_swagger -m -o --user carlosmart626@gmail.com swagger_2.yml

Cannot resolve keyword 'username' into field. Choices are: ....
```

PR #486, opened 2019-10-30
Fix issue #485

CarlosMart626:
With this update you can define custom models and custom user field and be used calling the command `generate_swagger` with `--user` option. Eg: `python manage.py generate_swagger --user carlosmart626@gmail.com`


CarlosMart626:
@axnsan12 do you have any comments on this change?

axnsan12:
Nope, seems good, thanks!

CarlosMart626:
@axnsan12 Thanks for merging this change! Do you think is possible to release a new version with this change?

PR #490, opened 2019-11-11
Edit type check for swagger_auto_schema (specifically for the `responses` param)

dlemfh:
Hello,

I use PyCharm and the `swagger_auto_schema` util has been causing **type-check** related warnings like the following *(see yellow box)*

image
(*sorry for lame screenshot..*)

Although the convention seems to be to use *integer* keys for the `responses` dict to pass to `swagger_auto_schema` ([reference](https://drf-yasg.readthedocs.io/en/stable/custom_spec.html#the-swagger-auto-schema-decorator)), type-checking was being done to enforce keys to be `str` in the code.

So I edited to allow both `int or str` types for the keys of the `responses` dict.

Not an important issue, but the warning in PyCharm was getting pretty annoying, so I made this PR..
Hope this  helps!


soulwawa:
hi, I'm still a problem.
help plz

스크린샷 2020-11-20 오후 4 32 35

PR #494, opened 2019-11-13
Add example for using swagger_schema_fields for a Field

### LINKED ISSUE #478: Provide an example to documentation for custom `Serializer Meta nested class`
Currently as of 0.17.0, the docs section https://drf-yasg.readthedocs.io/en/stable/custom_spec.html#serializer-meta-nested-class could really use an example.

For example, it's not immediately clear what should go into the `swagger_schema_fields` dictionary and how that will be translated into the resulting schema. An example (perhaps with a simple, nested `Schema`) would go a long way.

johnthagen:
Closes #478


johnthagen:
@etene Mind reviewing this PR?

etene:
Not at all, I'll take a look. At first glance, from a git history standpoint, wouldn't it be better to rebase your branch on master instead of including a merge commit ?

docs/custom_spec.rst
```diff
@@ -236,6 +236,62 @@ The available options are:
...
+.. Warning::
+
+   Overriding a default ``Field`` generated by a ``ModelSerializer`` will also override automatically
+   generated validators for that ``Field``. You may wish to add validation to your ``Serializer`` back manually using
+   the ``default_validators`` attribute on a field.
+
+   .. code-block:: python
+
+      class EmailMessageField(serializers.JSONField):
+         default_validators = [my_custom_email_validator]
```
etene: I think you should mention that it's not the only way to do it, and maybe point to the relevant documentation (see below) so you don't have to explain everything, which would be outside the scope of drf-yasg's documentation.

https://www.django-rest-framework.org/api-guide/validators/
https://www.django-rest-framework.org/api-guide/fields/#validators

docs/custom_spec.rst
```diff
@@ -236,6 +236,62 @@ The available options are:
      which are converted to Swagger ``Schema`` attribute names according to :func:`.make_swagger_name`.
      Attribute names and values must conform to the `OpenAPI 2.0 specification <https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#schemaObject>`_.
 
+Suppose you wanted model an Email this way, using a ``JSONField`` for performance reasons:
```
etene: I would rephrase that as 
> Suppose you wanted to model an email using a ``JSONField`` to store the subject and body for performance reasons:

johnthagen:
> Not at all, I'll take a look. At first glance, from a git history standpoint, wouldn't it be better to rebase your branch on master instead of including a merge commit ?

I was presuming that @axnsan12 would use GitHub's Squash Merge functionality to render that moot.

johnthagen:
@etene I believe I have addressed all of your review comments. Let me know what you think.

johnthagen:
Build failures appear unrelated?

axnsan12:
Yeah, seems good. Thanks!

PR #500, opened 2019-11-25
Describe responses with the docstring of their serializer

PaulWay:
If the response schema is being built from a serializer, then we can
use the serializer's docstring to document the meaning of the
response.

Signed-off-by: Paul Wayper <paulway@redhat.com>


michi88:
I think you wanted to make a pr in the new fork https://github.com/JoelLefkowitz/drf-yasg and not here, right?

PaulWay:
Hence https://github.com/JoelLefkowitz/drf-yasg/pull/48 :smile:

JoelLefkowitz:
Complete PR in #786

PR #507, opened 2019-12-06
Allow working with coreschema subclasses

alexkiro:
I'm currently working with some custom coreschema for my APIs by sub-classing the existing ones. As I'm migrating to OpenAPI (the coreschema compat was a life saver :heart:), found that drf-yasg doesn't catch these cases and falls back to strings. 

This should help out for any cases where subclasses are used for coreschema definitions.

### DIFF
--- a/src/drf_yasg/inspectors/query.py
+++ b/src/drf_yasg/inspectors/query.py
@@ -38,21 +38,27 @@ class CoreAPICompatInspector(PaginatorInspector, FilterInspector):
             'form': openapi.IN_FORM,
             'body': openapi.IN_FORM,
         }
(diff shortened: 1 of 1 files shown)


axnsan12:
Thanks for the contribution and sorry for ignoring this for so long. 

coreschema is as of now deprecated, and I'm going to drop the compat support, so this is obsolete. Closing.

alexkiro:
> Thanks for the contribution and sorry for ignoring this for so long.
> 
> coreschema is as of now deprecated, and I'm going to drop the compat support, so this is obsolete. Closing.

Understandable, no worries! :+1:

PR #542, opened 2020-02-11
Change wget to curl in update-ui.sh

rafalolszewski94:
Helps to use `update-ui.sh` on MacOS (wget is not present on Mac's)


rafalolszewski94:
Those checks above are not valid for this PR since I didn't change anything in `.py` files. It was bad formatted before.

axnsan12:
Seems good, curl is probably more commonplace.

PR #566, opened 2020-03-28
Fix x-nullable on serializers with `allow_null=True`

terencehonles:
This PR allows `x-nullable` on serializers. As originally implemented in #217 I believe this was too strict. The `required` `instance_kwargs` in this context are the fields that are required *not* if the field is required or not. According to the openapi spec the parent should be managing if the child is required and therefore this shouldn't even be checking `field.required` either.


qweeze:
Any chance of merging this any time soon?

terencehonles:
> Any chance of merging this any time soon?

I think @axnsan12 has either been burnt out or is affected by everything that's going on (directly or indirectly), since they haven't done anything with this repo since Feb

axnsan12:
Makes sense to me, thanks.

PR #572, opened 2020-04-04
Do not use field label for top level definition titles

terencehonles:
Along with #566 this PR allows use by tools such as https://github.com/Yelp/bravado/ which uses the title for the "model" name. Although their choice is slightly debatable, the titles should not be defined by the first serializer they are used by (What currently will happen). This change removes the titles if they are based on the field name and leaves the possibility they might be added in some other way in the future.


axnsan12:
This makes sense, especially if it improves usability with bravado. I imagine the title field is generally pretty uselss anyway. Thanks!

PR #621, opened 2020-07-30
Properly handle subclasses of SerializerMethodField

belak:
I can add a test if needed, but this is a fairly simple fix. There's a check for None after this, so this was clearly the intended behavior.

In particular, I think this shows up if you subclass SerializerMethodField and don't use method_name.


axnsan12:
Seems good, thanks.

PR #630, opened 2020-08-24
Documentation: Include openapi Import Example

Rjak:
Developers new to drf-yasg will benefit from seeing an example of how to import openapi.


axnsan12:
Makes sense, thanks

Issue #627, opened 2020-08-20
add support for primitive `list`


tomwojcik:
I tried to add `drf-yasg` to a rather complicated project with a lot of typing but it breaks on list.

I think the problem lies somewhere around here
https://github.com/axnsan12/drf-yasg/blob/9ccf24c27ad46db4f566170553d49d687b6d21d6/src/drf_yasg/inspectors/field.py#L481-L508

I'm not sure how exactly it is handled here but I don't think `list` should ever reach `inspect_collection_hint_class`.

Is there a reason for not having `(list, (openapi.ARRAY_FIELD, None))` in `hinting_type_info` ? 
Is it because of `typing.Sequence` that's being added? But `typing.List` has `__args__` and `list` doesn't.

Here it lacks of test case for `list`
https://github.com/axnsan12/drf-yasg/blob/9ccf24c27ad46db4f566170553d49d687b6d21d6/tests/test_get_basic_type_info_from_hint.py#L15-L40

Happy to open a PR with test and fix.

MacOS
Py 3.8
Django 3.1
DRF 3.11.1

Stacktrace

```
fs_django  | [INFO/fs:82] GET /swagger/?format=openapi REQUEST
fs_django  | [ERROR/root:15] Bad request
fs_django  | Traceback (most recent call last):
fs_django  |   File "/usr/local/lib/python3.8/site-packages/rest_framework/views.py", line 502, in dispatch
fs_django  |     response = handler(request, *args, **kwargs)
fs_django  |   File "/usr/local/lib/python3.8/site-packages/drf_yasg/views.py", line 94, in get
fs_django  |     schema = generator.get_schema(request, self.public)
fs_django  |   File "/usr/local/lib/python3.8/site-packages/drf_yasg/generators.py", line 254, in get_schema
fs_django  |     paths, prefix = self.get_paths(endpoints, components, request, public)
fs_django  |   File "/usr/local/lib/python3.8/site-packages/drf_yasg/generators.py", line 412, in get_paths
fs_django  |     operation = self.get_operation(view, path, prefix, method, components, request)
fs_django  |   File "/usr/local/lib/python3.8/site-packages/drf_yasg/generators.py", line 454, in get_operation
fs_django  |     operation = view_inspector.get_operation(operation_keys)
fs_django  |   File "/usr/local/lib/python3.8/site-packages/drf_yasg/inspectors/view.py", line 45, in get_operation
fs_django  |     responses = self.get_responses()
[...]

PR #631, opened 2020-08-29
add support for primitive list

tomwojcik:
Ticket  #627 

I don't like this `if` either but I have no idea how to fix it differently as you check for subclass for both primitives and `typing` objects.


tomwojcik:
I think my solution won't work with the upcoming 3.9 changes as `dict` is not supported either in this solution. 

https://docs.python.org/3.9/whatsnew/3.9.html#pep-585-builtin-generic-types

axnsan12:
Yes, the issue was due to 3.9 generics. Thanks for the report

PR #635, opened 2020-09-09
feat(decorators): let swagger_serializer_method accept many argument

souldeux:
Add an optional `many` kwarg to the `swagger_serializer_method` decorator.


souldeux:
oh lol you can do `@swagger_serializer_method(MySerializerClass(many=True))`

PR #645, opened 2020-09-29
Updated readme (added staticfiles in installed apps as a dependency)

iNishant:
added `django.contrib.staticfiles` in installed apps as a dependency. Required for serving swagger ui's css/js files


axnsan12:
Makes sense, thanks.

PR #646, opened 2020-09-30
Default to "/accounts/logout/" rather than settings.LOGOUT_URL

vanschelven:
settings.LOGOUT_URL hasn't been included in Django since Django 1.10,
and hasn't functioned since pre-1.0.

The default of "/accounts/logout/" corresponds to most examples in the
documentation, and is more likely to work out of the box than the
present default of `None`.

See https://github.com/django/django/commit/59f861fcb49a43acfbf2f75e683a7cb971600f52


axnsan12:
Thanks!

PR #648, opened 2020-10-01
Attempt patch compatibility with DRF 3.12.x

iflare3g:
…on command related to missing oauth2_provider migration

- Skip collecting one failing test due to an error by importing Token proxy model of DRF auth token app, known bug of DRF 3.12.x


tests/test_form_parameters.py
```diff
@@ -1,44 +1,45 @@
-import pytest
```
iflare3g, line 1: Let me know how to handle this test, it's failing due to the bug I highlighted into the conversation comments.
and I wanna avoid commenting it but I've tried with pytest.mark.xfail but the test raises an error and the marker seems not to work and I don't know how to mark it to be skipped instead of commenting it.

eremeevfd: @pytest.mark.skip - will skip the test

iflare3g: I've tried it as I used to do but the `pytest -n auto --cov ` still raises error for it

iflare3g: I will try again, my idea was to use pytest.mark.skipif and check the DRF version, because until 3.11.x, the test will pass

eremeevfd: Nice idea

iflare3g: I've found a solution and I will push it in a while

iflare3g: Just pushed the fix and updated docs.
Now pytest won't collect the test if DRF version is greater than 3.11.x. the pytest.mark.skip does not work if there's an error during the collection process.

ivlevdenis:
@axnsan12 what about merge this fix? On more projects this bug stopping use new versions of drf(

JoelLefkowitz:
> @axnsan12 what about merge this fix? On more projects this bug stopping use new versions of drf(

The maintainers of this repo are not replying, we've moved to my fork [drf-yasg2](https://github.com/JoelLefkowitz/drf-yasg) to patch issues and continue development

LanceMoe (approved):

iflare3g:
Fortunately my team is not blocked but it's weird that we cannot have the latest version of DRF due to this issue 😟

kingbuzzman:
We need this too! My Team is blocked with DRF 3.12 because of this. Whats the timeframe for a merge looking like?

terencehonles:
> We need this too! My Team is blocked with DRF 3.12 because of this. Whats the timeframe for a merge looking like?

It's unlikely anything will be merged anytime soon. You may want to create a fork of the repo and install with `pip install -e git+git://github.com/<your-org>/drf-yasg.git@master#egg=drf_yasg` this is usually what I have done when things were actually moving forward, but it's likely something your team can do also. You can specify this in a requirements.txt file by starting at the `-e` so... `-e git+git://github.com/<your-org>/drf-yasg.git@master#egg=drf_yasg`

JoelLefkowitz:
> We need this too! My Team is blocked with DRF 3.12 because of this. Whats the timeframe for a merge looking like?

Released today.
pip install dry-yasg2==1.18.0

axnsan12:
Implemented in master, dropped support for django <2.2, drf<3.10, thanks for investigating and sorry for not responding.

Issue #389, opened 2019-06-20
coreapi is deprecated


sad-boy-uoa:
Hi there,

I've just noticed that this library depends on the `coreapi` which has been deprecated. What is the strategy going forward to work around this?

As per the `coreapi` github page:

```
DEPRECATION NOTICE: As of djangorestframework 3.9, the plan is to phase out CoreAPI in favor of OpenAPI as the default schema representation. New projects should consider using OpenAPI rather than CoreAPI. See the DRF 3.9 release announcment for more details.
```

axnsan12:
Well, to be honest, if the OpenAPI implementation in DRF is up to snuff, the plan would be to deprecate this library altogether 🙂 Time will tell.

Otherwise, we'll have to review which parts of coreapi are used here and how we can replace them.

onegreyonewhite:
@axnsan12 You cant drop coreapi until in rest_framework.schemas.corepai.SchemaGenerator is used and DRF<3.11. But you can stop using coreapi imports. Check #588 for it.

johnthagen:
Another benefit of removing the dependency on `coreapi` is that `coreapi` depends on `requests`, which depends on `chardet`. `chardet` is LGPL licensed which presents problems for developers wanted to package their application as a binary using PyInstaller or PyOxidizer (https://github.com/psf/requests/issues/4417).

johnthagen:
Migrating issue details from https://github.com/JoelLefkowitz/drf-yasg/issues/60. Given that this is a long-term maintenance risk for the project, I am going to pin this issue for more visibility.

From the [`core-api` repo](https://github.com/core-api/core-api):

> DEPRECATION NOTICE: As of djangorestframework 3.9, the plan is to phase out CoreAPI in favor of OpenAPI as the default schema representation. New projects should consider using OpenAPI rather than CoreAPI. See the DRF 3.9 release announcment for more details.

[DRF will soon be raising deprecation warnings for CoreAPI schemas and they will be removed in DRF 3.14.](https://github.com/encode/django-rest-framework/issues/7007) I haven't looked closely enough at `drf-yasg2` to know if this will affect this package, or if we only rely on `coreapi` directly rather than *through* DRF itself.

[...]

PR #664, opened 2020-10-26
Remove direct usages of coreapi dependency

johnthagen:
First part of #389.  Remove direct usages of `coreapi` compatibility shims.

`coreschema` is left for another PR.

This PR does not fully remove `coreapi` dependency, because the `OpenAPISchemaGenerator` depends on `rest_framework.schemas.SchemaGenerator`, which still requires `coreapi`.

Note that I believe `rest_framework.schemas.SchemaGenerator` will be removed in DRF 3.14, so we need to eventually address this.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -1,3 +1,4 @@
+coreapi>=2.3.3
 coreschema>=0.0.4
 ruamel.yaml>=0.15.34
 inflection>=0.3.1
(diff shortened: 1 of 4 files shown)


johnthagen:
@axnsan12 @JoelLefkowitz Would you mind looking this over? As @axnsan12 said in #389, `coreapi` seems to be fairly easy to remove.

axnsan12:
Awesome. 

Travis status seems broken for some reason but I reset it and I think it will work on future PRs.

PR #665, opened 2020-10-28
Update Travis CI config to Python 3.9 final

johnthagen:


axnsan12:
Well, that was fast. I had tested just 2 days before and 3.9 was still 404. Awesome!

Issue #518, opened 2019-12-24
Type stubs for Mypy/PyCharm


intgr:
HI! I started a https://github.com/intgr/drf-yasg-stubs repository since drf-yasg was the only major component in my project that did not include type stubs. For now it's still quite incomplete (some of it is still auto-generated files with `Any` types).

I'm wondering what are your opinions on type stubs? [PEP 561](https://www.python.org/dev/peps/pep-0561/#specification) offers three ways to distribute stubs.

Option 3 is what I'm doing right now, with a separate package.

Clearly option 1 is out of the window if you want to retain Python 2 support. Although given its support ends ending in one week (and that Django already dropped it ages ago), perhaps dropping Python 2 altogether is due?

Option 2 would be to merge the `.pyi` files that I've written into this drf-yasg repository. This might be the reasonable first step if you don't want to drop Python 2 yet.

I'm willing to submit pull requests for it and do any other work that you deem necessary.

intgr:
Ping @axnsan12! I'm willing to put effort into finishing this work, just looking for some guidance and your thoughts.

intgr:
I have published my stubs package on PyPI: https://pypi.org/project/drf-yasg-stubs/

axnsan12:
Hello, if you're still interested in doing a PR for this, Python 2 support is going out the window soon. So adding inline type hints as annotations would be a nice to have.

intgr:
Hi! Thanks. Yes, I'm interested in doing that. Hopefully I will get around to it this week.

johnthagen:
Now that Python 2.7 has been dropped, this becomes a possibly now. It would be a good addition to the package in my opinion.

yoav-orca:
@johnthagen That's a great idea! happy to see this repo alive again

johnthagen:
Pinning this issue for more visibility, since I think it would be a good addition to the project.

We could also then add `mypy` checking to the CI.

intgr:
Thanks, I still would like to work on this but I have been busy with many things. Meanwhile drf-yasg-stubs should work fine.

Sorry if I failed anyone's expectations. But if someone else doesn't do it first, I will probably get to it eventually.

PR #666, opened 2020-10-29
Use a list, not a tuple, for permission_classes

brianhelba:
DRF expects permission classes to be in a list:
https://www.django-rest-framework.org/api-guide/permissions

Semantically, this should be a list, as the number of elements is arbitrary.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -148,7 +148,7 @@ In ``urls.py``:
          license=openapi.License(name="BSD License"),
       ),
       public=True,
-      permission_classes=(permissions.AllowAny,),
+      permission_classes=[permissions.AllowAny],
    )
 
    urlpatterns = [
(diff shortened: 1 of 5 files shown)


johnthagen:
Just a note, `tuple`s can semantically be arbitrarily long, but after creation they are immutable, unlike `list`s. You can see this in the type system with a type hint such as `Tuple[str, ...]`.

brianhelba:
@johnthagen Agreed, functionally, either can be used (`types.Sequence` would probably be best, if this used type annotations). However, `list` is documented by DRF and is more appropriate for use in this sort of case, where particular element locations do not have semantic meaning (as opposed to something like "RGB").

axnsan12:
Makes sense to match the DRF examples/docs here. In #518 we'd probably use Sequence.

PR #667, opened 2020-10-29
Do not use a leading dot to specify SwaggerRenderer formats

brianhelba:
DRF expects that custom Renderers use a simple string for the `format`: https://www.django-rest-framework.org/api-guide/renderers/#custom-renderers

This is also required to support content negotiation via DRF's native mechanisms, particularly `format_suffix_patterns`: https://www.django-rest-framework.org/api-guide/format-suffixes/

Note that all the other DRF-YASG renderers (including `OpenAPIRenderer` and `SwaggerUIRenderer`) already do the correct thing. This change just updates the `SwaggerJSONRenderer` and `SwaggerYAMLRenderer` to be consistent and correct.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -152,7 +152,7 @@ In ``urls.py``:
    )
 
    urlpatterns = [
-      url(r'^swagger(?P<format>\.json|\.yaml)$', schema_view.without_ui(cache_timeout=0), name='schema-json'),
+      url(r'^swagger\.(?P<format>json|yaml)$', schema_view.without_ui(cache_timeout=0), name='schema-json'),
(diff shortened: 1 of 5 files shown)


brianhelba:
Unfortunately, this is not compatible with the existing URL configuration that DRF-YASG's docs have been instructing users to configure. If this is released as-is, it would be a breaking change.

I'm open to ways to working in a compatibility mechanism, if it's agreed that this is a positive change.

JoelLefkowitz:
@creyD i think it's best to roll everything out compatibly and keep the public interfaces consistent with what the bulk of the user base are expecting. To support this url pattern we should extend the customisation support.

JoelLefkowitz:
Fixed in #911

PR #668, opened 2020-10-29
Use a list, not a tuple, for authentication_classes

brianhelba:
DRF expects authentication classes to be in a list: https://www.django-rest-framework.org/api-guide/authentication/

Semantically, this should be a list, as the number of elements is arbitrary.


brianhelba:
This has a minor conflict with #666, since they touch adjacent lines. If accepted, please merge #666 first. I'll rebase this to fix any conflicts afterwards.

Issue #248, opened 2018-11-06
[Help Needed] 500 Error when visiting API doc


Saad-K:
Hi,

I am getting an error when visiting the swagger api site, I have looked at the existing and old issues, searched around the internets but couldn't find an answer. I have also looked at the server logs, but I feel like I am missing something very obvious.
```
Failed to load API
Fetch errorInternal Server Error http://localhost/docs/?format=openapi
```
In my server logs I get the following message 
```
Internal Server Error: /swagger/
  | Traceback (most recent call last):
  |   File "/usr/local/lib/python3.6/site-packages/django/core/handlers/exception.py", line 34, in inner
  |     response = get_response(request)
  |   File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py", line 156, in _get_response
  |     response = self.process_exception_by_middleware(e, request)
  |   File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py", line 154, in _get_response
  |     response = response.render()
  |   File "/usr/local/lib/python3.6/site-packages/django/template/response.py", line 106, in render
  |     self.content = self.rendered_content
  |   File "/usr/local/lib/python3.6/site-packages/rest_framework/response.py", line 72, in rendered_content
  |     ret = renderer.render(self.data, accepted_media_type, context)
  |   File "/usr/local/lib/python3.6/site-packages/drf_yasg/renderers.py", line 34, in render
  |     return codec.encode(data)
  |   File "/usr/local/lib/python3.6/site-packages/drf_yasg/codecs.py", line 74, in encode
  |     return force_bytes(self._dump_dict(spec))
  |   File "/usr/local/lib/python3.6/site-packages/drf_yasg/codecs.py", line 104, in _dump_dict
  |     return json.dumps(spec)
  |   File "/usr/local/lib/python3.6/json/__init__.py", line 231, in dumps
  |     return _default_encoder.encode(obj)
  |   File "/usr/local/lib/python3.6/json/encoder.py", line 199, in encode
  |     chunks = self.iterencode(o, _one_shot=True)
  |   File "/usr/local/lib/python3.6/json/encoder.py", line 257, in iterencode
  |     return _iterencode(o, 0)
  |   File "/usr/local/lib/python3.6/json/encoder.py", line 180, in default
  |     o.__class__.__name__)
  | TypeError: Object of type 'Africa/Abidjan' is not JSON serializable
[...]

PR #717, opened 2021-05-11
add support for working with pytz object field

MilanPecov:
When a pytz object is serialized it fails with this error 

`TypeError: Object of type Africa/Abidjan is not JSON serializable` and this issue has already been reported (for example https://github.com/axnsan12/drf-yasg/issues/248)

The current workaround is to modify the methods in the pytz serializer field, but it becomes more complicated when other third party libraries in the project are also broken because of this.

So I think it is reasonable to assume that because any bigger real project will require working with a timezone, it's better to fix the problem here instead of making modifications on various fields and libraries. And it's just a few lines fix. 

Also I'm not sure if this change requires tests and if they are needed where should I add them?

Thanks


MilanPecov:
I would also label this a bug because when a pytz object is serialized the docs are not working at all

JoelLefkowitz:
@MilanPecov I agree this is good idea. If the user doesn't want the pytz objects to be parsed to strings they still have the option to modify the serializer. If there are other common field types that are limiting json encoding then we may need a separate method to handle them.

Issue #739, opened 2021-09-18
Migrate from Travis CI to Github actions


JoelLefkowitz:
The Travis CI pipeline is no longer running for free. I would like to switch over to github actions so that we can have PRs automatically checked. I'm happy to add the configuration files. Does anyone have other suggestions before I go ahead?

NeOneSoft:
I'm very agree with you @JoelLefkowitz since in my humble opinion git hub actions is already a great CI tool for git repos!!

JoelLefkowitz:
Linked PR #741

PR #415, opened 2019-07-17
Improve testing

Jamim:
Hello,

I hope you'll find this PR useful.

These changes:
 - Fix the `lint` and the `djmaster` jobs
 - Fix compatibility with upcoming Django 3.0
 - Replace `jobs` with `matrix` in `.travis.yml`
 - Add a test job for Python 3.8
 - Allow running tests on any branch

Best regards!

PR #741, opened 2021-09-20
Actions

JoelLefkowitz:
As mentioned in #739 Travis CI is no longer running CI for free. This PR allows Github actions to run unit tests. This will allow us to continue approving PRs.

The tox environments for linters, djmaster and docs have been bound to explicit python versions rather than rely on CI environment variables for this.

The djmaster testenv until now has been allowed to fail. This is now made explicit with a tox flag. It is still a useful check as it offers diagnostic information about future compatibility.


JoelLefkowitz:
I would like approval from at least one more frequent contributor before changing over the required checks settings.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,31 @@
...
+    name: Publish the package on pypi
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the source code
+        uses: actions/checkout@v2
+
+      - name: Set the python version
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.6
```
johnthagen, line 19: Is the idea that we publish using the older supported Python version? It might be worth a comment if so, for future people looking at this.

.travis.yml
```diff
@@ -1,70 +0,0 @@
-language: python
```
johnthagen, line 1: Could we also swap out the build status badge in `README.rst` for GitHub actions badge(s)?

https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/adding-a-workflow-status-badge

johnthagen:
@JoelLefkowitz I left a few comments, but in general **thank you** for putting the time into this. This looks great.

tox.ini
```diff
@@ -28,8 +33,8 @@ deps =
 
     # test with the latest builds of Django and django-rest-framework
     # to get early warning of compatibility issues
-    djmaster: https://github.com/django/django/archive/master.tar.gz
-    djmaster: https://github.com/ottoyiu/django-cors-headers/archive/master.tar.gz
+    djmaster: https://github.com/django/django/archive/main.tar.gz
+    djmaster: https://github.com/ottoyiu/django-cors-headers/archive/main.tar.gz
```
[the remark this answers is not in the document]
JoelLefkowitz: Looks like it was added #415 to provide the dependency for each django version there. Somewhere along the line those were removed except its listing under djmaster.  I agree with you it is superfluous since the djmaster test is only diagnostic anyway.

.github/workflows/publish.yml
JoelLefkowitz: Added in the next commit.

.travis.yml
JoelLefkowitz: Added in the next commit.

tox.ini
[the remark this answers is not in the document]
JoelLefkowitz: Removed in the next commit.

tox.ini
```diff
@@ -7,13 +7,18 @@ isolated_build_env = .package
 envlist =
     py{36,37,38,39}-django{22,30}-drf{310,311,312},
     py{36,37,38,39}-django31-drf{311,312},
```
johnthagen, line 9: Could we add Django 3.2 testing while we are here?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,36 @@
...
+        run: pip install -r requirements/publish.txt
+
+      - name: Build the distributions
+        run: python setup.py sdist bdist_wheel
+
+      - name: Publish the package
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          user: __token__
+          password: ${{ secrets.PYPI_API_TOKEN }}
```
johnthagen, line 36: Add newline at end of this file to make it consistent with the others?

johnthagen (approved):

JoelLefkowitz:
I think only @axnsan12 has permission to remove the travis status check and set the github actions ones as required.

johnthagen:
In the case that @axnsan12 has this repo muted (understandable given he's mentioned he's moved on to other things), his email address can be viewed in the commit log: https://github.com/axnsan12/drf-yasg/commit/a9ec562196a697c2fbfa16ad741f53831b8ab654.patch

Might be a way to get into contact with him.

axnsan12:
Hello, I do indeed have the repo muted 😢 but I do check on the notifications from time to time.

From what I see Travis still offers CI credits for open source projects: https://docs.travis-ci.com/user/billing-faq/#what-if-i-am-building-open-source

I could write to them to request this for drf-yasg and see how it goes. 

Or, if you'd prefer to switch to actions, I can remove the Travis check and be done with it.

johnthagen:
@axnsan12 I personally think that GitHub Actions is the way forward. Most open source projects I've worked on have switched, and it's been a net benefit for the project (faster CI, the marketplace helps avoid boilerplate, etc.).

axnsan12:
I hope you don't mind if I force merge this and see what happens.

johnthagen:
> The actions checks, like all checks that run on PRs, are required by default.

I don't think this is correct. The checks will show up in the PR, but if they are not required, then someone with merge rights to the repo can still merge even if the GitHub Actions checks fail.

axnsan12:
I've also set up the PYPI_API_TOKEN secret required by the publish job.

johnthagen:
Given @JoelLefkowitz's commitment to the project, for what it's worth I would nominate he be given admin access if @axnsan12 concurs.

axnsan12:
I'm not sure that's possible on a personal repo, we'd have to make an organisation to be able to have multiple owners 

Personal repos can only add "collaborators" - which both d you are.

johnthagen:
@JoelLefkowitz Since you were talking about cutting a new release, it probably wouldn't hurt to also update the bundled swagger-ui / redoc that is included in this project:

- https://github.com/axnsan12/drf-yasg/tree/master/src/drf_yasg/static/drf-yasg

https://github.com/axnsan12/drf-yasg/blob/effc8f2c06e66813a0ab5bad5a4478c0a8e1969c/package.json#L4-L5

tfranzel:
@axnsan12 @johnthagen, I had to migrate from **travis-ci.org** to **travis-ci.com** a while ago, because the builds simply stopped working. 

They do provide credits for OSS, but it's a painful process. You have to write a support ticket and and apply for OSS credits. That works fairly fast, but they only credit X amount at a time. Since migrating, I had to re-request credit 4-5 times. Multiple questions regarding a recurring "monthly allowance" were ignored.

axnsan12:
Thanks for the info, but we decided to just ditch Travis in the end. Seems like that was the right choice, your experience sure does sound like a lot of hassle 😄

PR #742, opened 2021-09-24
handle errors rendering with TemplateHTMLRenderer

terencehonles:
This change fixes an `OPTIONS` request to the API page which will return a successful response, but the `TemplateHTMLRender` will not be prepared to render it since [there is no template name provided in either the view or response](https://github.com/encode/django-rest-framework/blob/f0a5b958a134e8cd94e3ef3263e8fa623ac9b82f/rest_framework/renderers.py#L178-L189).


src/drf_yasg/renderers.py
```diff
@@ -64,10 +65,15 @@ class _UIRenderer(BaseRenderer):
...
-            # in that case, it's probably better to let the default ``TemplateHTMLRenderer`` render it
-            # see https://github.com/axnsan12/drf-yasg/issues/58
-            return TemplateHTMLRenderer().render(swagger, accepted_media_type, renderer_context)
+            try:
+                # if `swagger` is not a ``Swagger`` object, it means we somehow got a non-success ``Response``
+                # in that case, it's probably better to let the default ``TemplateHTMLRenderer`` render it
+                # see https://github.com/axnsan12/drf-yasg/issues/58
+                return TemplateHTMLRenderer().render(swagger, accepted_media_type, renderer_context)
+            except ImproperlyConfigured:
+                # Fall back to using eg '404 Not Found'
```
terencehonles, line 74: Stolen from the error handling code path https://github.com/encode/django-rest-framework/blob/f0a5b958a134e8cd94e3ef3263e8fa623ac9b82f/rest_framework/renderers.py#L199-L202

PR #749, opened 2021-11-02
Change url() to re_path()

by-gramm:
`url()` function is deprecated since Django 3.1.

Since drf-yasg is compatible with Django 3.1, I think it would be better to replace `url()` with `re_path()`.


axnsan12:
Fair. Thanks!

PR #758, opened 2021-12-10
Support new X | Y union syntax of Python 3.10 (PEP 604)

intgr:
This is used, for example, when inspecting the result type of `SerializerMethodField`.


creyD:
@intgr Did you test it back then properly after your initial comments?

intgr:
Heh this is awkward. I was actually using drf-spectacular, but submitted a PR for the wrong project. So I never finished this. 

I have rebased and pushed my latest changes, but the tests are still failing, so someone else should finalize this.

intgr:
I already added some tests, but they fail when running with Python 3.10. So some additional changes are needed to the logic. I don't know where it goes wrong.

Issue #756, opened 2021-11-27
License and source status unclear for some minimized files


tijuca:
# Bug Report

## Description
The upstream source of drf-yasg ships some minimized JS files where the source and by this the license isn't fully clear.

The greater background:
I packaged drf-yasg within Debian as [drf-yasg-nonfree](https://tracker.debian.org/pkg/drf-yasg-nonfree). It's placed within the non-free part (which isn't officially Debian due non-free) as some file are minimized and no source for these files nor a minimal information how to rebuild the minimized files can be rebuilt are added. The ftp-masters want to have a more clear state about the files where the license and source they are taken from isn't fully clear.

In detail these are the *.js files in [the folder src/drf_yasg/static/drf-yasg](https://github.com/axnsan12/drf-yasg/tree/master/src/drf_yasg/static/drf-yasg) and beyond.

So far I could a do research I found out the following things.

[src/drf_yasg/static/drf-yasg/immutable.min.js ](https://github.com/axnsan12/drf-yasg/blob/master/src/drf_yasg/static/drf-yasg/immutable.min.js) is based on https://github.com/immutable-js/immutable-js/releases/tag/v3.8.2

[src/drf_yasg/static/drf-yasg/url-polyfill.min.js](https://github.com/axnsan12/drf-yasg/blob/master/src/drf_yasg/static/drf-yasg/url-polyfill.min.js) is based on https://github.com/lifaon74/url-polyfill

[src/drf_yasg/static/drf-yasg/insQ.min.js](https://github.com/axnsan12/drf-yasg/blob/master/src/drf_yasg/static/drf-yasg/insQ.min.js) no source can be found, license is readable from the minimized file.

[src/drf_yasg/static/drf-yasg/redoc-init.js](https://github.com/axnsan12/drf-yasg/blob/master/src/drf_yasg/static/drf-yasg/redoc-init.js) no source can be found, I assume the source is taken from https://github.com/Redocly/redoc/ or the minimized file is somehow taken from there. By this it's assumed the license would be Expat as this is the license of GitHub tree.

[src/drf_yasg/static/drf-yasg/swagger-ui-init.js](https://github.com/axnsan12/drf-yasg/blob/master/src/drf_yasg/static/drf-yasg/swagger-ui-init.js) Partially the same as before, but it's mostly obvious the source must be taken from https://github.com/swagger-api

[...]

PR #759, opened 2021-12-12
Adding source files for existing minimized files in src/drf_yasg/static/drf-yasg/

tijuca:
Based on discussion in #756 I hereby request adding the current non existing source files for the minimized JS files in the folder `src/drf_yasg/static/drf-yasg/`. Adding these files shouldn't hurt but makes it more easy for downstream to package this Python library.

Beside adding the non minimized JS files I added some legal information to the existing files `redoc-init.js` and `swagger-ui-init.js` so it's clear what license and copyright holders are taking effect.


axnsan12:
Thanks!

PR #786, opened 2022-04-08
Add descriptions to fields and responses

PaulWay:
This was added to drf-yasg2 a while back and didn't get backported into drf-yasg.


JoelLefkowitz:
@PaulWay thanks for the time put into this. I've corrected the implementation and updated some of the test schema. 

At the moment the unit tests are not functionally testing behaviour, rather they match the test output to a correct test schema. Since I've updated the 'correct' schema to include descriptions I'd like some additional reviewers to verify the changes before merging. More specifically, are the definitions that are now in `tests/reference.yaml` as they should be?

src/drf_yasg/inspectors/field.py
```diff
@@ -105,14 +129,25 @@ def make_schema_definition(serializer=field):
...
+                    )
+                
+                # Don't set a description if it's an empty string or None
+                else:
+                    result = SwaggerType(
+                        use_field_title=False,
+                        type=openapi.TYPE_OBJECT,
+                        properties=properties,
+                        required=required or None
+                    )
```
onegreyonewhite, line 150: @PaulWay SwaggerType is, in fact, `drf_yasg.openapi.Schema`. Maybe it's better to insert description separately?

```python
if description:
    result.description = description
```

Another way:
```python
result = SwaggerType(
    ...
    description=description or None
)
```

src/drf_yasg/inspectors/field.py
```diff
@@ -557,7 +592,7 @@ def field_to_swagger_object(self, field, swagger_object_type, use_references, **
         if not isinstance(field, serializers.SerializerMethodField):
             return NotHandled
 
-        method = getattr(field.parent, field.method_name, None)
+        method = getattr(field.parent, field.method_name)
```
onegreyonewhite, line 595: Why? Without default value this code will raise AttributeError.

src/drf_yasg/inspectors/field.py
```diff
@@ -72,15 +77,34 @@ def _has_ref_name(self, serializer):
 
     def field_to_swagger_object(self, field, swagger_object_type, use_references, **kwargs):
         SwaggerType, ChildSwaggerType = self._get_partial_types(field, swagger_object_type, use_references, **kwargs)
+        
+        description = getattr(field, 'description', getattr(field, '__doc__'))
+        
+        if description is None and hasattr(field, 'Meta') and hasattr(field.Meta, 'model') and hasattr(field.Meta.model, '__doc__'):
+            description = field.Meta.model.__doc__
```
onegreyonewhite, line 84: This doesn't seem like a good idea, because the Model is an internal entity and may contain help that may not be relevant to the API at all.

onegreyonewhite (commented):
@PaulWay good job! I think these changes will help the schema to be more understandable for developers.

JoelLefkowitz:
@PaulWay I've cherry picked your two commits because the branch had diverged

JoelLefkowitz:
Thanks for the suggestions on the PR @onegreyonewhite,

I've put the `getattr(..., None)` back and removed this section: 

```py
if description is None and hasattr(field, 'Meta') and hasattr(field.Meta, 'model') and hasattr(field.Meta.model, '__doc__'):
    description = field.Meta.model.__doc__
```

PaulWay:
Hi @JoelLefkowitz - yeah, I was wading through some ugly git rebase work there and I'm sure I was getting it wrong.  Cherry-picking those commits is a much better plan :-)

src/drf_yasg/utils.py
```diff
@@ -516,3 +516,14 @@ def get_field_default(field):
...
     return default
+
+
+def strip_doc_string(doc):
+    """
+    Remove line breaks and excess space from the start and end of each line of
+    a docstring. Ignores indenting, so some formatting will be lost.
+    """
+    if not doc:
+        doc = ""
```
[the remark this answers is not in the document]
PaulWay: Thanks Joel and it's definitely just a personal thing rather than trying to find any consistent style in the drf-yasg coding :-)

I use the early return idiom for situations where there's an obvious 'null case' like this, where no further processing needs to occur.  It's mainly for readability, but it's also better performance for the null case.  I tested this and the `doc=""` code runs in 0.37 seconds in the `doc=None`, whereas the `return ""` code runs in 0.04 seconds.  Both cases run in 0.48 seconds when `doc="Foo bar baz"`.

And stylistically it definitely wins over:

```py
if not doc:
    return ""
else:
    return " ".join(line.strip() for line in doc.splitlines() if line.strip())
```
Why have the else and the indent there, indeed?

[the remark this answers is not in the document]
PaulWay: Your return type argument is a good one and for larger functions it would be more of a consideration.

PR #735, opened 2021-09-01
Add support for Django 3.2

NeOneSoft:
This PR was created in order to add support for Django 3.2

PR #803, opened 2022-08-03
Declare Django 3.2 support in README, classifiers

DavidCain:
This commit will make sure that `Django :: 3.2` will show up in the
classifiers list on PyPI: https://pypi.org/project/drf-yasg/

image

(The magic happens because we parse the README to get supported Django):
https://github.com/DavidCain/drf-yasg/blob/ee29412d3cdb311/setup.py#L36

3.2 support should already exist
================================
A closed pull request, https://github.com/axnsan12/drf-yasg/pull/735,
noted support for Django 3.2 in both `tox.ini`, and the README.

That PR was closed in favor of https://github.com/axnsan12/drf-yasg/pull/741, which edited `tox.ini` and switched to GitHub Actions.

https://github.com/axnsan12/drf-yasg/pull/735#issuecomment-924298555

This project has been testing on Django 3.2 for a long time (about a
year). I think we can declare it supported it the README!

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -12,7 +12,7 @@ Generate **real** Swagger/OpenAPI 2.0 specifications from a Django Rest Framewor
 Compatible with
 
 - **Django Rest Framework**: 3.10, 3.11, 3.12
-- **Django**: 2.2, 3.0, 3.1
+- **Django**: 2.2, 3.0, 3.1, 3.2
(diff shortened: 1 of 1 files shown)


sowinski:
Is this project dead or will Django 3.2 support released soon?

Issue #765, opened 2022-01-04
ruamel.yaml 0.17.0 has deprecated dump(), load()


jwhitlock:
# Bug Report

``drf_yasg`` is using functions that have been deprecated in ruamel.yaml 0.17.0, released 2021-03-26:

> the old top level *functions* `load`, `safe_load`, `round_trip_load`, `dump`, `safe_dump`, `round_trip_dump`, `scan`, `parse`, `compose`, `emit`, `serialize` as well as their `_all` variants for multi-document streams, now issue a `PendingDeprecationning` (e.g. when run from pytest, but also Python is started with `-Wd`). Use the methods on `YAML()`, which have been extended.

I believe this affects only ``yaml_sane_dump`` and ``yaml_sane_load``.

## Is this a regression?

Not a ``drf-yasg`` one. ruamel.yaml 0.16.3, the first supported version and the last to support Python 2, does not have this deprecation warning. It also may not have the extended arguments to `YAML()`..

## Minimal Reproduction
I discovered this while testing the endpoint with pytest in https://github.com/mozilla/fx-private-relay/pull/1447.

Another way to reproduce it is with a short bit of code. Add this to a working Django / drf-yasg project as `repo_bug.py`:
```python
from django.conf import settings

def test_round_trip():
    from drf_yasg.codecs import yaml_sane_dump, yaml_sane_load

    data = {"key": "value"}

    out = yaml_sane_dump(data, False)
    round_trip = yaml_sane_load(out)
    assert round_trip["key"] == "value"

if __name__ == "__main__":
    settings.configure()
    test_round_trip()
```

Run with ``python -Wd repo_bug.py``. The ``-Wd`` is needed to display the ``PendingDeprecationWarning``s.

## Stack trace / Error message

```code
../lib/python3.7/site-packages/drf_yasg/codecs.py:202: PendingDeprecationWarning: 
dump will be removed, use

  yaml=YAML(typ='unsafe', pure=True)
  yaml.dump(...)

instead
  return yaml.dump(data, Dumper=SaneYamlDumper, default_flow_style=False, encoding='utf-8' if binary else None)
../lib/python3.7/site-packages/ruamel/yaml/main.py:1380: PendingDeprecationWarning: 
dump_all will be removed, use

  yaml=YAML(typ='unsafe', pure=True)
  yaml.dump_all(...)

instead
  block_seq_indent=block_seq_indent,
[...]

PR #806, opened 2022-08-25
Fix 765 - Update to ruamel.yaml 0.17, use new API

jwhitlock:
This PR reimplements YAML loading and unloading using the `ruamel.yaml` API introduced in 0.15. The older PyYAML-based API is deprecated in 0.17. This fixes issue #765 and the deprecation warnings.

Much of the YAML code was in the initial commit in 2017. Since then, `ruamel.yaml` has implemented fixes and features that removes the need for a lot of the custom code.

I tested by:
* Running the test suite with `tox`
* Running the `testproj` app before and after the change. The YAML is identical.


jwhitlock:
It looks like the project switched from `ruamel.yaml` to `pyyaml`, this is no longer relevant.

JoelLefkowitz:
Thanks @jwhitlock :)

PR #807, opened 2022-08-25
Fix documentation generation

jwhitlock:
This allow `tox -e docs` to run without warnings or errors. This was mostly fixing class references in docstrings. A few other changes:

* Switch `license_file` to `license_files` in setup.cfg. This was added to setuptools in v42.0.0 (Nov 2019), and `license_file` was deprecated in v57.0.0 (May 2021).
* Fix the `envlist` for `tox`. Running `tox -e py38-lint` would run pytest, not flake8, and `tox -e py38-docs` would not build docs.


JoelLefkowitz:
Yup, the docs now build without any warnings, thanks you for the time you put into the PR @jwhitlock :)

PR #499, opened 2019-11-22
Response schemas from serializers will optionally be paginated

PaulWay:
At the moment, if a non-default response serializer is set via `responses` in `swagger_auto_schema`, it will not pick up any pagination schema that might apply from any paginator inspectors.

This patch will take any serializers provided in the `responses` section and apply a pagination schema to them if the view should be paged.  This does not apply to the default serializer from `get_default_responses` because that is already converted into a schema.

Signed-off-by: Paul Wayper <paulway@redhat.com>

PR #809, opened 2022-08-31
Update view.py

lopezvit:
This relays on https://github.com/axnsan12/drf-yasg/pull/730 and https://github.com/axnsan12/drf-yasg/pull/499 but with the newest code, as I need to implement it in my own project.


JoelLefkowitz:
@lopezvit could you please rebase this PR?

Could you also please explain this segment in `view.py`:

```py
if self.has_list_response():
    schema = openapi.Schema(type=openapi.TYPE_ARRAY, items=schema)

if self.should_page():
    schema = self.get_paginated_response(schema) or schema
```

It looks like it comes from `get_default_responses`. Could we call `get_default_responses` instead?

Issue #810, opened 2022-09-22
Compatibility issues with djangorestframework==3.14.0


matthew-hodgins:
# Bug Report

djangorestframework upgraded to version 3.14.0 last night. This breaks compatibility with drf-yasg 1.21.3.

## Description

With version 3.14.0, rest_framework.serializers.NullBooleanField was removed. 

## Is this a regression?

No.

## Minimal Reproduction

```code

```

## Stack trace / Error message

```code
  File "/home/.../lib/python3.6/site-packages/drf_yasg/inspectors/field.py", line 406, in <module>
    (serializers.NullBooleanField, (openapi.TYPE_BOOLEAN, None)),
AttributeError: module 'rest_framework.serializers' has no attribute 'NullBooleanField'
```

## Your Environment

```code
Python 3.6.9
Django==3.2.15
djangorestframework==3.14.0
drf-yasg==1.21.3
```

na0495:
i did encounter the same issue with the env
```
Django==3.2.9
djangorestframework>=3.12.2
drf-yasg>=1.20.0
```

pablodiegoss:
Same issue here, had to pin djangorestframework to 3.13.1 to work with drf-yasg 1.21.3

JoelLefkowitz:
Thanks for flagging, Ill get a fix out for this today.

PR #814, opened 2022-09-24
Fix compatibility with django-rest-framework 3.14.0 having depreciated NullBooleanField

JoelLefkowitz:
As pointed out in #810, django-rest-framework 3.14.0 has depreciated its NullBooleanField. 

For the sake of backwards compatibility we can drop the NullBooleanField types when the installed version of django-rest-framework exceeds 3.14.0.

### DIFF
--- a/src/drf_yasg/inspectors/field.py
+++ b/src/drf_yasg/inspectors/field.py
@@ -42,7 +42,7 @@ class InlineSerializerInspector(SerializerInspector):
         is called only when the serializer is converted into a list of parameters for use in a form data request.
 
         :param serializer: serializer instance
-        :param list[openapi.Parameter] parameters: genereated parameters
+        :param list[openapi.Parameter] parameters: generated parameters
         :return: modified parameters
         :rtype: list[openapi.Parameter]
         """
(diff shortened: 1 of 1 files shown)


JoelLefkowitz:
Before merging I would like someone to have a look at this PR for a soundness check. Once merged I'll deploy a new release to pypi.

yamac-kurtulus:
Just a heads up, models.NullBooleanField is still there, serializer field is deprecated. Though it is completely understandable if you want to remove that too.

src/drf_yasg/inspectors/field.py
```diff
@@ -390,7 +394,7 @@ def decimal_field_type(field):
     (models.TimeField, (openapi.TYPE_STRING, None)),
     (models.UUIDField, (openapi.TYPE_STRING, openapi.FORMAT_UUID)),
     (models.CharField, (openapi.TYPE_STRING, None)),
-]
+]
```
sondrelg, line 397:
```suggestion
]
```

JorisBenschop:
indeed, at this moment when a user installs the latest drf and  drf-yasg, there is a crash in "serializer_field_to_basic_type" (field.py).

PR #818, opened 2022-10-15
Run tests with latest dependencies

nikosmichas:
I wanted to use this library in a new project and I noticed that tests do not run with latest Django & DRF versions

- [x] Added entries in tox for:

- Django 4.0 & 4.1
- DRF 3.13 & 3.14
- Python 3.10

- [x] Updated psycopg2-binary in test requirements to run with Python 3.10

- [x] Fixed an AttributeError raised because of a deprecation in DRF


StevenMapes:
From my own uses of this project is works on Python up to 3.10 with Django up to 4.1. It is not Python 3.11 compatible due to ruamel-yaml-clib not supporting Python 3.11. Issue #833 suggests moving to PyYAML instead which is the default YAML serializer for Django which would probably resolve that issue as well

JoelLefkowitz:
@nikosmichas, thanks for the PR. Im changing the base to 1.21.x so the tests can run with latest set of release changes. The NullBooleanField changes have been handled in #814 and are present there too.

nikosmichas:
Tests are failing because github no longer supports py36 for GH actions. I will take a look on this.

JoelLefkowitz:
> Tests are failing because github no longer supports py36 for GH actions. I will take a look on this.

Thanks @nikosmichas, this is now fixed in the last commit

Issue #613, opened 2020-07-09
RecursiveField schema generation is not working


mbrancato:
It looks like the support for `RecursiveField` added in #109 is no longer working. The Swagger UI won't load and presents an error. I noticed there is no test for this field type in the `/tests/` folder as well - I'd help create one but I'm having some difficulty getting all the requirements to install.

relevant versions:
```
djangorestframework==3.9.3
djangorestframework-recursive==0.1.2
drf-yasg==1.17.1
```

Using a simple example like this:

```python
class TestResponseSerializer(serializers.Serializer):
    parent = RecursiveField()

@swagger_auto_schema(
    query_serializer=TestRequestSerializer,
    responses={200: TestResponseSerializer},
)
@action(detail=False, methods=["get"])
def test(self, request):
    ...
```

The error is the following:
```
AssertionError: #/definitions/TestResponse is not defined
```

axnsan12:
Interesting that this slipped by without tests. Would be nice to fix.

JoelLefkowitz:
I'll take a look into this

aprilahijriyan:
Any progress here?

JoelLefkowitz:
Fixed in #822

PR #822, opened 2022-10-25
Fix `RecursiveField`

iorlandini:
Fix `RecursiveField` as pointed out broken in https://github.com/axnsan12/drf-yasg/issues/613.


JoelLefkowitz:
@iorlandini thanks for catching this

PR #824, opened 2022-11-29
Hide read_only=True nested serializers from data

curiousKonstantin:
1. Use "allOf" OpenAPI property to combine $ref object with extra parameters with $ref object
2. Hide definition from request payload representation if read_only=True argument was provided for a nested serializer
3. Show "description" for a $ref if "help_text" was provided for a nested serializer


curiousKonstantin:
I was annoyed by frontenders constantly ask "Why do you expect an object in payload?" and was frustrated by impossibility to show help_text from nested serializer on swagger schema (because each field meaning need to be explained, separately) so I've made this workaround with allOf property to fix the issues.

Issue #817, opened 2022-10-13
Support for the latest versions of Django and DRF.


Ayub-Khan:
# Feature Request
Support for the latest versions of Django and DRF.

## Description
I wanted to create this thread to track Django 4 support work progress. I could provide a senior resource to speed up the process if we could get a gist of what exactly needs to be done for this.

## Describe the solution you'd like
The resource I provide won't have much knowledge of the code base, so it will probably be a code overview meeting. and guidance on how you would like that to progress and track. 

## Describe alternatives you've considered
--

StevenMapes:
For reference for anyone coming to this I'm currently using this with Django 4.2 and Django 3.14.0 running Python 3.11.2. I've also run it on Django 3.2, 4.0 and 4.1 so it the compatibility list needs to be updated to show the full support

Issue #771, opened 2022-02-07
Python 3.10 Support


JackLeo:
# Feature Request

## Description

`Setup.py` reads `README.rst` and limits python compatibility up to 3.9. DRF does support Python 3.10. 
Is there a particular reason or list of bugs why 3.10 is not supported?

## Describe the solution you'd like

Update `README.rst` to list Python 3.10

## Describe alternatives you've considered

N/A

JackLeo:
Apologies if spam. Search did not result in anything in the issues/PRs related to this, just DRF 3.10 version support.

JoelLefkowitz:
Thanks for the suggestion @JackLeo, #758 contains some of the changes for python3.10. The tox targets and GitHub actions matrix will be updated to include python3.10 once support for all its features have been implemented.

PR #836, opened 2023-02-09
Adding Python 3.11 into the test matrix

StevenMapes:
I've added Python 3.11 into the test matrix and README.rst as I am using the package on Python 3.11 and have personally run it with both 3.11.1 and 3.11.2 as a followup to #771 .

I'll test Django 4.2-alpha at some-point as well as I'm pretty sure it will work which will go alongside #817

### DIFF
--- a/.github/workflows/review.yml
+++ b/.github/workflows/review.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: "ubuntu-20.04"
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


onegreyonewhite:
@StevenMapes already done with #837.  Might need to be closed to keep order.

StevenMapes:
> @StevenMapes already done with #837. Might need to be closed to keep order.

Okay I see your later issue/PR has been merged so I'll close this and tag it in a comment on that for reference

PR #840, opened 2023-03-03
fix type hints when using postponed evaluation of annotations (PEP-563)

terencehonles:
This change uses `typing.get_type_hints` instead of `inspect.signature` since the former will try to `eval` string type hints (PEP-563) and will return an object rather than a string if it was successful which will allow the remaining code to work. [`inspect.signature`](https://docs.python.org/3/library/inspect.html#inspect.signature) on the other hand does not have the ability to resolve string type hints until Python 3.10 and has to be called with `eval_str=True`

This also partially reverts https://github.com/axnsan12/drf-yasg/pull/428, but fixes https://github.com/axnsan12/drf-yasg/issues/426 correctly by not failing on the `isclass(hint_class)` check due to annotations not "really" being classes.

This latter change is done because `list[str]` is unconditionally and improperly unwrapped and will result in a type of `str` since `list`'s annotations will *also* have a first argument of `str`.


terencehonles:
Marking this to be merged after #868, but combining both PRs to confirm this fixes the lint check

terencehonles:
This change fixes when running with `from __future__ import annotations` and also handles the Python 3.10 syntax and pulls in the test from https://github.com/axnsan12/drf-yasg/pull/758 to ensure it does

terencehonles:
Thanks for the other merge @JoelLefkowitz , I've rebased this and hopefully it's clear what it's doing. It removes some helper methods that are only used once which makes it clearer what's happening at the call site(s) and also allows some optimizations by hoisting some work out of the loop.

PR #841, opened 2023-03-03
keep path parameters in their given order

terencehonles:
This change keeps path parameters in their given order as tools like https://github.com/swagger-api/swagger-codegen/ use the parameter ordering as is to create a function signature. Arguably the tools could re-parse the path looking for template variables and use that ordering but it appears they do not.

The swagger spec does not seem like it documents how these parameters should be ordered, and while when reading the schema document, alphabetically ordered might make sense for a human to verify the parameters exist, it's losing information about the ordering which may be conveying additional information.

### DIFF
--- a/src/drf_yasg/generators.py
+++ b/src/drf_yasg/generators.py
@@ -495,7 +495,7 @@ class OpenAPISchemaGenerator(object):
         parameters = []
         queryset = get_queryset_from_view(view_cls)
 
-        for variable in sorted(uritemplate.variables(path)):
+        for variable in uritemplate.variables(path):
             model, model_field = get_queryset_field(queryset, variable)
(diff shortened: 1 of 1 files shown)


riddhesh6si:
Hi @terencehonles , @JoelLefkowitz 
This PR messes up any function that was called using positional arguments instead of keyword arguments as it is completely changing the ordering. 

> it's losing information about the ordering which may be conveying additional information.

 what additional information might be getting conveyed by the order of params ?

terencehonles:
@riddheshSajwan this change should fix that issue not creates it. Before the parameters were sorted, but after removing the sorting they will be in the order that they are supplied in the path, which is the order that tools like Django will call a view with.

Using tools like https://github.com/swagger-api/swagger-codegen relied on this order to create API clients which now had the incorrect order. Is this maybe what you are suggesting (that a generated client's order is now changed)?

PR #849, opened 2023-05-18
Added support for zoneinfo objects

nikvst:
Same fix as https://github.com/axnsan12/drf-yasg/pull/717, but for [zoneinfo](https://docs.python.org/3/library/zoneinfo.html) objects.

I ran into a problem because django [is migrating from pytz to ZoneInfo](https://docs.djangoproject.com/en/4.2/releases/4.0/#zoneinfo-default-timezone-implementation) and is planning to remove pytz support from django 5.0.

If this change requires tests, where should I add them?

Thanks


src/drf_yasg/utils.py
```diff
@@ -1,6 +1,7 @@
 import inspect
 import logging
 import textwrap
+import zoneinfo
```
yuekui, line 4: `zoneinfo` was introduced from Python 3.9, shall we consider checking Python version when importing it?

nikvst: Oh, you're right, thanks. I fixed it.

jgrewal1991:
@axnsan12 when can we expect this to merge?

JoelLefkowitz:
@jgrewal1991 there are currently merge conflicts. We would need someone to resolve these first and to check the CI workflows are passes too

jgrewal1991:
> @jgrewal1991 there are currently merge conflicts. We would need someone to resolve these first and to check the CI workflows are passing too

I can give it a shot. @JoelLefkowitz

JoelLefkowitz:
@jgrewal1991 Thank you

JoelLefkowitz (approved):

nikvst:
Hello @JoelLefkowitz 

I can pull the latest master and make a new commit to avoid resolving conflicts. I can do this over the weekend.
Would that help?

JoelLefkowitz:
@nikvst thank you for the original work :)

PR #853, opened 2023-06-12
Upgrade deprecated pkg_resources

AngelOnFira:
Locally, I'm running tests and getting deprecation warnings for the use of `pkg_resources`. According to [this post](https://setuptools.pypa.io/en/latest/pkg_resources.html):

---

Use of pkg_resources is deprecated in favor of [importlib.resources](https://docs.python.org/3/library/importlib.resources.html#module-importlib.resources), [importlib.metadata](https://docs.python.org/3/library/importlib.metadata.html#module-importlib.metadata) and their backports ([importlib_resources](https://pypi.org/project/importlib_resources), [importlib_metadata](https://pypi.org/project/importlib_metadata)). Users should refrain from new usage of pkg_resources and should work to port to importlib-based solutions.

---

This PR makes the changes to follow the deprecation notice.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -12,7 +12,7 @@ import sphinx_rtd_theme
 from docutils import nodes, utils
 from docutils.parsers.rst import roles
 from docutils.parsers.rst.roles import set_classes
-from pkg_resources import get_distribution
+from importlib.metadata import version
 
 # -- General configuration ------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


AngelOnFira:
Note; this package uses CoreApi, and it also uses `pkg_resources`. However, that repo is archived, so I can't upstream a PR there.

tahme:
Note that `importlib.metadata` is only available for Python >= 3.8 so older versions need to use the backported `importlib_metadata` package or keep using `pkg_resources`, like here:
https://github.com/pyscaffold/pyscaffold/pull/318/files#diff-832c6d0841c720713101f1313509d4255dacb74d3cc2c58f05cbe52c53de53ee

cclauss:
* #891

JoelLefkowitz:
Closing this as its fixed in #891 as @cclauss pointed out

PR #859, opened 2023-06-29
[fix] Fixed map source mapping

pandafy:
Closes #855

PR #860, opened 2023-07-05
Update redoc.min.js sourceMapURL to point to correct map path

anthonyalmarza:
When running `manage.py collectstatic` on a project that also uses [whitenoise](https://whitenoise.readthedocs.io/en/latest/index.html) automated builds will break due to an incorrect `sourceMapURL` path. The following error is reported.

```shell
whitenoise.storage.MissingFileError: The file 'drf-yasg/redoc-old/redoc.min.map' could not be found with <whitenoise.storage.CompressedManifestStaticFilesStorage object at 0x7f9f03d5ed10>.

The JS file 'drf-yasg/redoc-old/redoc.min.js' references a file which could not be found:
  drf-yasg/redoc-old/redoc.min.map

Please check the URL references in this JS file, particularly any
relative paths which might be pointing to the wrong location.
```

A workaround is to copy the related statics maintained here into your own project (fixing the path of course).


JoelLefkowitz:
Completed in #859 thanks @anthonyalmarza!

PR #868, opened 2023-09-07
fix lint errors with type comparison

terencehonles:
This fixes some lint errors that appeared when updating https://github.com/axnsan12/drf-yasg/pull/840 since the linting rules do not seem to be pinned or changed since the last run.


terencehonles:
I'll see if I can update this PR sometime soon, but may not get to it this weekend

terencehonles:
I've rebased this after #897  @JoelLefkowitz

terencehonles:
@JoelLefkowitz do you mind if we get this small PR in? They don't seem to be lint errors anymore, but they should be using `is` instead of `==`.

PR #873, opened 2023-11-04
Support Python 3.12

max-muoto:
`pkg_resources` is no longer accessible by default in Python 3.12, which means this library is incompatible with Python 3.12. This PR moves from using `pkg_resources` to `importlib.metadata` which is the intended way of replacing the metadata API of `pgk_resources`. See [here](https://docs.python.org/3/library/importlib.metadata.html#module-importlib.metadata). 

Since `pkg_resources` is only included in Python 3.8+, this PR also drops support for early versions of Python. If support for these older versions is something that we want to keep, I'm happy to go and add in conditional logic to use a backport for older versions to maintain compatibility, but as 3.7 is now in end-of-life I assumed we might want to drop support now.

Fixes https://github.com/axnsan12/drf-yasg/issues/874.


max-muoto:
@JoelLefkowitz Not sure who the best person to ping on this is, but would love to hear thoughts on this.

JoelLefkowitz:
@max-muoto thanks for the PR I've approved the unit test run and I'll have a look at the changes

max-muoto:
Thanks Joel, let me know if you think we need to make any changes here! Seeing the 3.12 run fail, but it looks like that might just be due to a cancellation.

*
*

*Maxwell Muoto*

***@***.***

650-704-1494

Sent via Superhuman ( ***@***.*** )

JoelLefkowitz:
@max-muoto Id like to keep the support for python 3.6 and 3.7 since a lot of our users are tied to those versions and compatibility is a priority. Could you add them back? Looks like the tox command is failing for python 3.12, could you try running it locally and see why? Thank you!

max-muoto:
> @max-muoto Id like to keep the support for python 3.6 and 3.7 since a lot of our users are tied to those versions and compatibility is a priority. Could you add them back? Looks like the tox command is failing for python 3.12, could you try running it locally and see why? Thank you!

Sounds good, I'll make sure we keep support there. Let me give it a try locally!

max-muoto:
@JoelLefkowitz 3.12 failures seem to be coming from usage of https://github.com/Yelp/swagger_spec_validator which appears to be unmaintained. Any ideas on potential workarounds here? Not sure how deep `drf-yasg` relies on this library.

max-muoto:
They have the same issue around `pgk_resources` usage.

JoelLefkowitz:
@max-muoto It says here setuptools isn't available by default anymore https://github.com/python/cpython/issues/95299. I've added it to the dependencies list and the python3.12 tests are now passing. I'm going to update the PR target to the release branch. Looks like flake8 is failing in the python 3.8 runs, could you take a look please?
Thank you

Morette:
can someone get back to this?
I just raised this bug that this PR apparently fix

https://github.com/axnsan12/drf-yasg/pull/873

max-muoto:
Closing this PR since I was able to find a workaround here, but if someone else has the time to take to get this out that would be great! Apologies here.

cclauss:
* #891

PR #885, opened 2024-05-13
Update utils.py

haradhansharma:
More straight forward parser consumes


cclauss:
Please rebase.

JoelLefkowitz:
Hey @haradhansharma could you please clarify the details of this commit?

The return type of the function is a list:

```
media_types = [parser.media_type for parser in parser_classes or []]
non_form_media_types = [encoding for encoding in media_types if not is_form_media_type(encoding)]

if len(non_form_media_types) == 0:
    return media_types

return non_form_media_types
```

But this expression may be an individual object:
```
return [media_types[0]]
```

Also `media_types` can be empty, so we can't index it without account for that.

I agree that we can drop the `or []` in the `media_types` expression because `parser_classes` is guaranteed to be a list

PR #891, opened 2024-10-10
Review: Add Python 3.11 and 3.12 to the testing

cclauss:
https://devguide.python.org/versions/

This is still Django v4 but not yet v5.

### DIFF
--- a/.github/workflows/review.yml
+++ b/.github/workflows/review.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python: ["3.6", "3.7", "3.8", "3.9", "3.10", "3.11"]
+        python: ["3.6", "3.7", "3.8", "3.9", "3.10", "3.11", "3.12"]
 
     steps:
       - name: Checkout the source code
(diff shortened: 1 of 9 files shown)


JoelLefkowitz:
Hey @cclauss thanks for raising this, could you please update the envlist in the tox.ini file too for the newer versions of python? Also we need to keep the version 3.6 run in the version matrix as it's required by the branch protection rules for this repo and cannot be removed without the original author of the repo.

cclauss:
Let's get the currently supported versions of Python and Django working before we circle back to unsupported versions.

.github/workflows/review.yml
```diff
@@ -26,23 +26,23 @@ jobs:
...
       - name: Report coverage
-        if: ${{ matrix.python == 3.9 }}
+        if: ${{ matrix.python != 3.6 }}
         run: |
           pip install coverage
           coverage report
 
       - name: Check for incompatibilities with publishing to PyPi
-        if: ${{ matrix.python == 3.8 }}
+        if: ${{ matrix.python == 3.11 }}
```
JoelLefkowitz, line 45: In the `publish.yml` workflow we use python3.8 to run twine so we should match these versions. Can you edit your PR to make the publish workflow use python3.11 too please?

.github/workflows/review.yml
```diff
@@ -26,23 +26,23 @@ jobs:
...
         # A successful workflow run for Python 3.6 is required by the GitHub branch protection rules.
         if: ${{ matrix.python != 3.6 }}
-        
+
         env:
           PYTHON_VERSION: ${{ matrix.python }}
         run: tox -e $(tox -l | grep py${PYTHON_VERSION//.} | paste -sd "," -)
 
       - name: Report coverage
-        if: ${{ matrix.python == 3.9 }}
```
JoelLefkowitz, line 39: Let's bump this up to python3.11 too if we can

JoelLefkowitz (changes requested):

.github/workflows/review.yml
cclauss: I believe this is correct as is.  `if: ${{ matrix.python != 3.6 }}` says that we are reporting `coverage` on all versions of Python except the Python 3.6 because `coverage` no longer supports Py36.

cclauss:
A lot of the changes in this PR are about supporting Python 3.7.  Should we stop supporting it and treat it like we do Py36?

Should we drop Py3.8 which is also EOL?

Should we drop Django < 4.2 which are also EOL?

.github/workflows/review.yml
JoelLefkowitz: You're right

JoelLefkowitz:
> A lot of the changes in this PR are about supporting Python 3.7. Should we stop supporting it and treat it like we do Py36?
> 
> Should we drop Py3.8 which is also EOL?
> 
> Should we drop Django < 4.2 which are also EOL?

Since we don't have any reliable feedback from the users on what versions they are using I'd like to keep supporting as many versions as we can. Since this is quite a mature package I'd like to prioritise stability

JoelLefkowitz (approved):

PR #893, opened 2024-10-14
GitHub Actions: review: strategy: fail-fast: false

cclauss:
When we test changes from Python 3.6 through 3.13 and Django 2.2 to 5.2, some versions will pass CI tests, and others will fail.  Give contributors more insight on which are which so they can focus on properly fixing the failures.

https://docs.github.com/en/enterprise-cloud@latest/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast


JoelLefkowitz:
@cclauss Thank you

PR #897, opened 2024-10-16
Merge release branches

JoelLefkowitz:
Our feature branching has led to branches getting old and drifting from the master branch. Initially we kept the master branch a few releases back but this doesn't provide any additional stability since we tag each release individually.


terencehonles:
It might have been better to create a merge commit instead of squashing everything, but should be simple enough to cherry pick things onto the main branch

PR #903, opened 2024-10-19
Lint Python code with ruff instead of flake8 and isort

cclauss:
https://docs.astral.sh/ruff
> Ruff can be used to replace [Flake8](https://pypi.org/project/flake8/) (plus dozens of plugins), [Black](https://github.com/psf/black), [isort](https://pypi.org/project/isort/), [pydocstyle](https://pypi.org/project/pydocstyle/), [pyupgrade](https://pypi.org/project/pyupgrade/), [autoflake](https://pypi.org/project/autoflake/), and more, all while executing tens or hundreds of times faster than any individual tool.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -6,10 +6,9 @@
 import inspect
 import os
 import re
-import sys
 import shutil
+import sys
 
-import sphinx_rtd_theme
 from docutils import nodes, utils
 from docutils.parsers.rst import roles
 from docutils.parsers.rst.roles import set_classes
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,39 @@
 [build-system]
-requires = ["setuptools >= 68.0.0", "wheel", "setuptools-scm >= 3.0.3"]
 build-backend = "setuptools.build_meta"
+
+requires = [ "setuptools>=68", "setuptools-scm>=3.0.3", "wheel" ]
+
+[tool.ruff]
+exclude = [ "**/migrations/*" ]
+line-length = 88
+lint.select = [
+  "AIR",   # Airflow
+  "ASYNC", # flake8-async
+  "C4",    # flake8-comprehensions
+  "C90",   # mccabe
+  "E",     # pycodestyle errors
+  "F",     # Pyflakes
+  "FA",    # flake8-future-annotations
+  "FLY",   # flynt
+  "I",     # isort
+  "ICN",   # flake8-import-conventions
+  "INT",   # flake8-gettext
+  "LOG",   # flake8-logging
+  "NPY",   # NumPy-specific rules
+  "PD",    # pandas-vet
+  "PLE",   # Pylint errors
+  "PYI",   # flake8-pyi
+  "SLOT",  # flake8-slots
+  "T10",   # flake8-debugger
+  "TCH",   # flake8-type-checking
+  "W",     # pycodestyle warnings
+  "YTT",   # flake8-2020
+]
+lint.ignore = [ "E501" ]
+lint.isort.known-first-party = [ "drf_yasg", "testproj", "articles", "people", "snippets", "todo", "users", "urlconfs" ]
+lint.mccabe.max-complexity = 13
+lint.per-file-ignores."src/drf_yasg/inspectors/field.py" = [ "E721" ]
+lint.per-file-ignores."src/drf_yasg/openapi.py" = [ "E721" ]
+lint.per-file-ignores."testproj/testproj/settings/*" = [ "F405" ]
+
+[tool.setuptools_scm]
--- a/requirements/lint.txt
+++ b/requirements/lint.txt
@@ -1,4 +1,2 @@
-# used by the 'lint' tox env for linting via flake8
-isort>=5.12
-flake8>=6.0.0
-flake8-isort>=6.0.0
+# used by the 'lint' tox env for linting via ruff
+ruff>=0.11.0
--- a/src/drf_yasg/codecs.py
+++ b/src/drf_yasg/codecs.py
@@ -3,8 +3,8 @@ import json
 import logging
 from collections import OrderedDict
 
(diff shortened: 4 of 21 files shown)


JoelLefkowitz:
This is interesting, I've never heard of ruff before. Speed for the static checkers hasn't been a concern for this repository before, but the simpler configuration and invocation of ruff looks really good

JoelLefkowitz:
I also like the idea of separating the linters from tox, since it's meant to be a testing tool

cclauss:
Many projects put subsecond tools like codespell, ruff, etc into https://pre-commit.com
Jupyter: https://github.com/jupyter/notebook/blob/main/.pre-commit-config.yaml

cclauss:
Scarry to run tests on Python 3.6 but not Python 3.13.

JoelLefkowitz:
> Scarry to run tests on Python 3.6 but not Python 3.13.

We've stopped the python3.6 tests a while ago, but I don't have enough permissions on the repo to remove it as a required check:

```yml
# Python 3.6 tests have been removed since swagger-spec-validator no longer supports it.
# A successful workflow run for Python 3.6 is required by the GitHub branch protection rules.
if: ${{ matrix.python != 3.6 }}
```

JoelLefkowitz:
Thanks for this PR @cclauss,

 I've had a look at the branch and would like to simplify the `pyproject.toml` a little:

```toml
[build-system]
build-backend = "setuptools.build_meta"
requires = ["setuptools>=68", "setuptools-scm>=3.0.3", "wheel"]

[tool.ruff]
exclude = ["**/migrations/*"]
line-length = 80

lint.select = [
    "ASYNC", # flake8-async
    "C4",    # flake8-comprehensions
    "C90",   # mccabe
    "E",     # pycodestyle
    "F",     # pyflakes
    "FA",    # flake8-future-annotations
    "FLY",   # flynt
    "I",     # isort
    "ICN",   # flake8-import-conventions
    "INT",   # flake8-gettext
    "LOG",   # flake8-logging
    "PLE",   # pylint
    "PYI",   # flake8-pyi
    "SLOT",  # flake8-slots
    "T10",   # flake8-debugger
    "TCH",   # flake8-type-checking
    "W",     # pycodestyle warnings
    "YTT",   # flake8-2020
]

lint.isort.known-first-party = [
    "articles",
    "drf_yasg",
    "people",
    "snippets",
    "testproj",
    "todo",
    "urlconfs",
    "users",
]

lint.mccabe.max-complexity = 13
lint.per-file-ignores."src/drf_yasg/inspectors/field.py" = ["E721"]
lint.per-file-ignores."src/drf_yasg/openapi.py" = ["E721"]
lint.per-file-ignores."testproj/testproj/settings/*" = ["F405"]
```

After this PR is merged I'd like to run the formatter in and create a new PR. The code is currently not compliant with any formatter. When we run `ruff format` most of the source folder will be modified. I've checked whether setting the quotes to 'single' would have less impact but it still updates about 50 files. The line width we choose comes down to opinion and I suggest we stick with the 80 characters limit that is the standard, even though I know lots of people don't like it.

Other related things that I'd like to do in separate PRs:
* Migrate the requirements files to the `pyproject.toml`'s `[project.optional-dependencies]`
* Remove the linters from the `tox.ini` since it should be responsible for testing only
* Add a linting step to the `review` workflow (for the latest python version only)

cclauss:
line-length = 80 instead of the default 88?

JoelLefkowitz:
The [line-too-long (E501)](https://docs.astral.sh/ruff/rules/line-too-long/#line-too-long-e501) rule suggests 88, but the [PEP 8](https://peps.python.org/pep-0008/#maximum-line-length) suggestion is 80. The ruff docs say they use 88 for compatibility with Black and the Ruff formatter. It's arbitrary which one we pick since they're both followed by large projects. I checked the Django repository and they're using 88. I think we should stick to the PEP 8 convention though.

cclauss:
Please read the justification for the change in psf/black docs.

JoelLefkowitz:
@cclauss is this the section you mean?

> You probably noticed the peculiar default line length. Black defaults to 88 characters per line, which happens to be 10% over 80. This number was found to produce significantly shorter files than sticking with 80 (the most popular), or even 79 (used by the standard library). In general, [90-ish seems like the wise choice](https://youtu.be/wf-BqAjZb8M?t=260).

Although shorter files is certainly a good thing I think that sticking to the most popular length is the smarter choice. My preference is 120 but I'd much rather just stick to PEP 8 because that aligns with what a majority of people are used to and we get a lot of contributors.

cclauss:
My point is that since black and ruff format have taken over the automated Python formatting space, MOST Python codebases are 88.  Checkout NumPy, SciPy, Django, FastAPI, Pydantic, etc.  I bet that in really popular codebases it will be difficult to find 80 char limits anymore.

JoelLefkowitz:
I found this [stack overflow answer](https://stackoverflow.com/a/2793525) pretty convincing when I researching what the standard is:

> Much of the value of PEP-8 is to stop people arguing about inconsequential formatting rules, and get on with writing good, consistently formatted code. Sure, no one really thinks that 79 is optimal, but there's no obvious gain in changing it to 99 or 119 or whatever your preferred line length is. I think the choices are these: follow the rule and find a worthwhile cause to battle for, or provide some data that demonstrates how readability and productivity vary with line length. The latter would be extremely interesting, and would have a good chance of changing people's minds I think.

cclauss:
It was written 15 years ago and does represent the way things were in that epoch.  Things have changed in the interim.

JoelLefkowitz:
Tbh I really don't mind as long as we stick to it. If it's the default for `ruff format` then lets keep 88, could you please update the PR?

JoelLefkowitz:
@cclauss could you please update the PR to 88?

JoelLefkowitz:
Thank you @cclauss

PR #911, opened 2025-03-07
update JSON & YAML renderers to not use a "." in their format string

terencehonles:
This is an updated version of https://github.com/axnsan12/drf-yasg/pull/667 which allows it to be backward compatible


JoelLefkowitz:
Thanks @terencehonles this is really great!

In #667 @brianhelba said:
> Unfortunately, this is not compatible with the existing URL configuration that DRF-YASG's docs have been instructing users to configure. If this is released as-is, it would be a breaking change.
> 
> I'm open to ways to working in a compatibility mechanism, if it's agreed that this is a positive change.

Could you explain how this approach resolved the compatible issue?

terencehonles:
Yes, this creates 2 additional renderers that also have `.json` and `.yaml` as the format if you don't set the setting `USE_COMPAT_RENDERERS = False` and those will be dispatched to if you don't change your urlconf. If you change the setting to `False` then they won't be created and your urlconf will need to match a format that isn't `r`\.(json|yaml)'`

StevenMapes:
Having just upgraded to the version that released this change in a project I think it would be a good idea to update both the README and the ReadTheDocs documentation to cover the SWAGGER_USE_COMPAT_RENDERERS setting and the new default setup so people can run without deprecation warning which will happen if you just follow the current install steps as ```SWAGGER_USE_COMPAT_RENDERERS=False``` is needed in order to avoid the DeprecationWarning which could break some CI setups

terencehonles:
Yeah that's a good suggestion, do you mind creating a PR @StevenMapes ?

PR #916, opened 2025-03-14
allow overriding produces/consumes with @swagger_auto_schema decorator

terencehonles:
This change allows overriding produces/consumes without having to subclass

### DIFF
--- a/src/drf_yasg/inspectors/view.py
+++ b/src/drf_yasg/inspectors/view.py
@@ -473,11 +473,13 @@ class SwaggerAutoSchema(ViewInspector):
 
         :rtype: list[str]
         """
-        return get_consumes(self.get_parser_classes())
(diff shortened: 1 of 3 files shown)


JoelLefkowitz:
Thanks @terencehonles, could you please add a unit test for this feature?

terencehonles:
@JoelLefkowitz done (and added a couple others while I am at it, since we don't test any of the overrides right now)

terencehonles:
Looks like Python 3.6 is dead :skull:, the last version on GA to support it is 20.04 https://github.com/actions/setup-python/issues/544#issuecomment-1332535877 and that's being pulled...

terencehonles:
@JoelLefkowitz I updated this to include the commit that fixed the CI

JoelLefkowitz:
Thanks @terencehonles 🥳

PR #917, opened 2025-03-14
fix list views with parameters in last path segment not named "list" views

terencehonles:
This fixes the auto generated operation keys and operation ID. For list views which have a parameter in their last path segment.

### DIFF
--- a/src/drf_yasg/utils.py
+++ b/src/drf_yasg/utils.py
@@ -315,18 +315,18 @@ def is_list_view(path, method, view):
     """
     # for ViewSets, it could be the default 'list' action, or an @action(detail=False)
     action = getattr(view, "action", "")
-    method = getattr(view, action, None) or method
-    detail = getattr(method, "detail", None)
+    detail = getattr(view, "detail", None)
     suffix = getattr(view, "suffix", None)
     if action in ("list", "create") or detail is False or suffix == "List":
         return True
 
     if (
         action in ("retrieve", "update", "partial_update", "destroy")
+        or method.lower() != "get"
+        # a detail action is surely not a list route
         or detail is True
         or suffix == "Instance"
     ):
-        # a detail action is surely not a list route
         return False
 
     if isinstance(view, ListModelMixin):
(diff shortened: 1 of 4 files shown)


src/drf_yasg/utils.py
```diff
@@ -227,8 +227,7 @@ def is_list_view(path, method, view):
     """
     # for ViewSets, it could be the default 'list' action, or an @action(detail=False)
     action = getattr(view, 'action', '')
-    method = getattr(view, action, None) or method
```
terencehonles, line 230: When looking at git blame the `or method` was added to silence the linter. I don't think `method` should be stomped on like this since it should be a string (also confirmed in the docstring), and `str.detail` will always be none.

terencehonles: Debugging this it looks like `view` has a `detail` for the test project's Article view set decorated with `@action`, and I think this was probably a typo, and instead of preserving `getattr(getattr(view, action, None), 'detail', None)` which won't work for the common case, I'm suggesting fixing this to `getattr(view, 'detail', None)` which for sure has a value.

JoelLefkowitz:
Thanks for digging into this @terencehonles, looks really good. I'd like to have a unit test for this to validate the behaviour when there is and isn't a detail attribute before we merge. Could we also move the `if method.lower() != "get":` clause into the previous if statement so it's a little easier to follow?

terencehonles:
> Thanks for digging into this @terencehonles, looks really good. I'd like to have a unit test for this to validate the behaviour when there is and isn't a detail attribute before we merge.

Isn't this what the reference schema is testing? I effectively fixed the test with https://github.com/axnsan12/drf-yasg/pull/917/files#diff-741bf237457101c6d297e385e111f8b5477fbbd9369c53a05c08c64e307d0e17

> Could we also move the `if method.lower() != "get":` clause into the previous if statement so it's a little easier to follow?

Done

terencehonles:
@JoelLefkowitz I updated this to include the commit that fixed the CI

PR #918, opened 2025-03-17
ruff format

cclauss:
https://docs.astral.sh/ruff/formatter


cclauss:
It would be possible to:
```toml
[tool.ruff.format]
quote-style = "single"
```
https://docs.astral.sh/ruff/formatter/#configuration

PR #920, opened 2025-03-18
disable ruff rules C90 and FA100

terencehonles:
The ruff rules added in https://github.com/axnsan12/drf-yasg/pull/918 are a bit aggressive for this project, and it's not clear how they were chosen. The re-formatting, while making my life harder in the short term, is a reasonable change.

There are many lint rules that are selected that probably don't make sense in this code base (i.e. "NPY" - numpy isn't even used in this repo), but I'm suggesting disabling FA100 since I don't believe we want to enable `from __future__ import annotations` as this _may_ be a breaking change to do automatically. For mccabe, it's not clear why 13 was chosen, and rather than picking an arbitrary number which will require function rewrites arbitrarily and without just cause, we just continue to let PR authors think about their code complexity and suggest refactoring during review.


terencehonles:
We can either follow up on this PR with enabling/disabling more rules, or follow up with other PRs.

cclauss:
Bad idea to remove limits to code complexity.  I would encourage keeping `lint.mccabe.max-complexity` in place and when new pull requests need to increase this value because they are making the code more difficult to understand and maintain then there should be a discussion in the code review about why the project needs more complexity and there is no way to refactor to limit that complexity.

cclauss:
> NPY Numpy is not used today...

Ruff runs in a small fraction of a second across the whole CPython codebase with no cache so adding a NPY is costless and if NymPy is ever added then the code is linted.

> For mccabe, it's not clear why 13 was chosen

Set it to 12 to reveal where code complexity exists.

> will require function rewrites arbitrarily and without just cause

Run `ruff rule C901` and read the `[Why is this bad?](https://docs.astral.sh/ruff/rules/complex-structure/#why-is-this-bad)` section.

terencehonles:
I'm sorry, but how much have you participated in this project already? I understand that code should be simplified, but this is already a very complex codebase and arbitrarily saying "this is too complex" is too naive at this point. We need to figure out how and where we should target code complexity and then work on getting things in a reasonable state to turn something like this on.

I also understand that ruff is fast, but you obviously are copying and pasting from somewhere and I'm just suggesting we be a bit more thoughtful about what we're trying to lint for.

terencehonles:
If there is a consensus that C90 should be turned on, what I propose is to **not** set it at 13 and keep it at the default of 10, and instead grandfather all offenders as allowed (add `# noqa: C90`). This is better because something that just happened to be at 12 and now is 13 doesn't require a refactoring by adding a simple change.

JoelLefkowitz:
Hey @terencehonles and @cclauss really interesting points here. There are a few different questions being discussed in one PR here:

_Should we use C90?_
Yes, too much complexity makes it harder for contributors.

_Why is the threshold set to 13?_
It's a common approach in large repositories to set the initial linter rules to be 'just passing' and then slowly bring them back to the defaults. Some functions like `get_schema_view` have complexity of 12 so it’s a good starting point. We can bring it down in future PRs one point at a time.

_Should we add `# noqa: C90` instead of setting the threshold above the default?_

> arbitrarily saying "this is too complex" is too naive at this point

This is a really good point, however I think it will be a lot simpler manage if we start the linter at a higher threshold rather than adding exceptions around the codebase.

_Should we include the NPY linter?_
The `airflow` and `pandas-vet` linters are not used here either. Originally it was much larger list but in #903 but we trimmed it down. I would like to remove all these linters that we don't use since they add noise to the `pyproject.toml` file. If we ever added numpy we could always add the extra linter rule again.

_Should we remove FA100?_

> I'm suggesting disabling FA100 since I don't believe we want to enable from `__future__` import annotations as this may be a breaking change to do automatically

FA100 doesn't allow `pyupgrade` to introduce `__future__` imports, rather it allows `pyupgrade` to remove the additional imports if the minimum python version doesn't require them anymore.

The next steps I would like use to take are:
* Keep FA100 and C90
* Keep the initial C90 threshold to 13
* Bring the C90 down to 10 in subsequent PRs
* Remove the AIR, NPY and PD linters rules
* Add the minimum python version for the FA linter rules

Does anyone feel strongly against this approach?

terencehonles:
> _Should we use C90?_ Yes, too much complexity makes it harder for contributors.
> 
> _Why is the threshold set to 13?_ It's a common approach in large repositories to set the initial linter rules to be 'just passing' and then slowly bring them back to the defaults. Some functions like `get_schema_view` have complexity of 12 so it’s a good starting point. We can bring it down in future PRs one point at a time.
> 
> _Should we add `# noqa: C90` instead of setting the threshold above the default?_
> 
> > arbitrarily saying "this is too complex" is too naive at this point
> 
> This is a really good point, however I think it will be a lot simpler manage if we start the linter at a higher threshold rather than adding exceptions around the codebase.

In order to give an opinion on this I'll need to look at the codebase a little (and see what's flagged when changing this)

> _Should we remove FA100?_
> 
> > I'm suggesting disabling FA100 since I don't believe we want to enable from `__future__` import annotations as this may be a breaking change to do automatically
> 
> FA100 doesn't allow `pyupgrade` to introduce `__future__` imports, rather it allows `pyupgrade` to remove the additional imports if the minimum python version doesn't require them anymore.

I'm not referring to pyupgrade, I'm referring to the ruff check, and I am not sure what you're saying is correct. Based on https://docs.astral.sh/ruff/rules/future-rewritable-type-annotation/ it says it tries to force you to use [PEP-563](https://peps.python.org/pep-0563/) if your annotation can be more simply expressed with it. I don't believe we want to use PEP-563 since it does change things if you're not already using it (may not really make a difference for this project), but PEP-563 is being rejected in favor of [PEP-649](https://peps.python.org/pep-0649/), but that will not land until Python 3.14, and we likely want to leave PEP-563 off until we can drop Python 3.13 support (a long time away).

The reason I am only aware of this new rule is because I was required to silence the linter in the following code:

https://github.com/axnsan12/drf-yasg/blob/cc02fccae408fb453ac407b53e032ba962d32d70/tests/test_schema_generator.py#L354-L355

I made the change since we don't need to support Python 2 anymore and can write the annotation correctly now.

Otherwise, I think the approach seems reasonable.

terencehonles:
I'll try to get to that next week, but for now it looks like we should focus on https://github.com/axnsan12/drf-yasg/pull/922 as it fixes the build, then https://github.com/axnsan12/drf-yasg/pull/916 because it's small and has been reviewed (test added), and finally https://github.com/axnsan12/drf-yasg/pull/917 because it should be ready or close to ready to merge.

JoelLefkowitz:
Thank you for the work that went into this @terencehonles

Issue #662, opened 2020-10-25
NOTICE: Consider using drf-yasg2 as an active fork of this project


johnthagen:
Unfortunately the community has not been able to contact @axnsan12 for many months. During this time, DRF 3.12 has been released, which is no longer compatible with `drf-yasg` (#641), as well as Django 3.1 and Python 3.9.

In order to continue to keep this project from dying, @JoelLefkowitz gratuitously volunteered to champion a new fork, `drf-yasg2`, which has allowed us to merge in some long awaited fixes (#625) as well as make sure everything is tested on the latest Python, Django, and DRF.

We want to thank @axnsan12 for all of his great work on this project and hope this notice helps other users find a working package for their projects.

axnsan12:
Hello there! 

First of all let me say that I'm very sorry for disappearing and abandoning this project. I won't bore you with excuses but suffice to say that I'm no longer an active user of drf-yasg, DRF, Django, or web development as a whole.

As such I won't have motivation for developing this project any further. 

I've just now been reading over the open issues and PRs I'm very happy to see that other people have taken up the intiative of forking and publishing an updated verison.

As for the fate of this repo, I'm going to do what needs to be done to get it running on the latest Django/drf, so that people who are still using the original package can have a functional version. I'm open to adding you @johnthagen  or @JoelLefkowitz, or anyone else interested, as maintainers to keep this alive.

JoelLefkowitz:
Hiya Cristi!
Glad to hear from you, sorry to hear you’ve left the wonderful world of web dev. 
Let us help you establish a functional version, we can merge in our fixes from the fork and adjust them if you see fit. 
Please do make us maintainers and we'll keep this repo nice and healthy

fasih:
I would also like to become a maintainer of this repo to keep it nice and healthy.

creyD:
So with the new maintainers, will this become the main version again? This would make this issue pretty much obsolete and we should merge the work done in this fork here and try to migrate it's users back here, right?

johnthagen:
Yes, I would say that when @JoelLefkowitz (given the initiative he's shown, I feel as though he makes the most sense as a second maintainer) is promoted to a maintainer, this issue can be closed and we can begin the work of backporting whatever we need from the fork.

johnthagen:
As `drf-yasg` is back in action, I am closing this issue. 🎉

johnthagen:
[...]

PR #922, opened 2025-03-19
update Python, Django, and DRF versions & packaging configuration

terencehonles:
This change is a proposal, it drops the EoL Python versions, and limits the Django support to the last two major versions (~4 years since release) and the DRF versions that support those versions.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,41 +1,30 @@
 [tox]
-minversion = 3.3.0
-isolated_build = true
-isolated_build_env = .package
+minversion = 4.0.0
 
 # https://docs.djangoproject.com/en/dev/faq/install/#what-python-version-can-i-use-with-django
 envlist =
-    py{37,38,39}-django{22,30}-drf{310,311,312},
-    py{37,38,39}-django{31,32}-drf{311,312},
-    py{39,310}-django{40,41}-drf{313,314},
-    py311-django{40,41,42,50,51}-drf315,
-    py312-django{42,50,51}-drf315,
-    py313-django{51}-drf315,
-    py38-{lint, docs},
-    # py313-djmaster  # ModuleNotFoundError: No module named 'psycopg'
+    py3{9,10,11,12   }-django40-drf{313,314,315},
+    py3{9,10,11,12,13}-django41-drf{313,314,315},
+    py3{9,10,11,12,13}-django42-drf{314,315},
+    py3{  10,11,12,13}-django50-drf{314,315},
+    py3{  10,11,12,13}-django5{1,2}-drf{315},
+    py3{        12,13}-prerelease,
+    py39-{lint,docs}
 
 skip_missing_interpreters = true
 
-[testenv:.package]
-# no additional dependencies besides PEP 517
-deps =
-
-[testenv:py310-djmaster]
+[testenv:py31{2,3}-prerelease]
 ignore_outcome = true
 
 [testenv]
 deps =
-    django22: Django>=2.2,<2.3
-    django30: Django>=3.0,<3.1
-    django31: Django>=3.1,<3.2
-    django32: Django>=3.2,<3.3
     django40: Django>=4.0,<4.1
     django41: Django>=4.1,<4.2
     django42: Django>=4.2,<5
+    django50: Django>=5.0,<5.1
+    django51: Django>=5.1,<5.2
+    django52: Django>=5.2,<6
 
-    drf310: djangorestframework>=3.10,<3.11
-    drf311: djangorestframework>=3.11,<3.12
-    drf312: djangorestframework>=3.12,<3.13
     drf313: djangorestframework>=3.13,<3.14
     drf314: djangorestframework>=3.14,<3.15
     drf315: djangorestframework>=3.15,<3.16
@@ -44,20 +33,16 @@ deps =
 
     # test with the latest builds of Django and django-rest-framework
     # to get early warning of compatibility issues
-    djmaster: https://github.com/django/django/archive/main.tar.gz
-    djmaster: https://github.com/encode/django-rest-framework/archive/master.tar.gz
+    prerelease: https://github.com/django/django/archive/main.tar.gz
+    prerelease: https://github.com/encode/django-rest-framework/archive/master.tar.gz
 
     # other dependencies
-    -r requirements/validation.txt
     -r requirements/test.txt
-    drf310: coreapi>=2.3.3
-    drf310: coreschema>=0.0.4
-    drf310: flex~=6.14.1
-
+extras = validation
 commands =
-    pytest -n 2 --cov --cov-config .coveragerc --cov-append --cov-report="" {posargs}
+    pytest -n auto --cov --cov-config .coveragerc --cov-append --cov-report="" {posargs}
 
-[testenv:py38-lint]
+[testenv:py39-lint]
 skip_install = true
 deps =
     -r requirements/lint.txt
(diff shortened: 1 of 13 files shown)


tox.ini
```diff
@@ -1,41 +1,40 @@
...
+    # Python 3.10 & 3.11: Django <= 5.2
+    py31{0,1}-django{40,41}-drf{313,314,315},
+    py31{0,1}-django{42,50}-drf{314,315},
+    py31{0,1}-django{51,52}-drf315,
+    # Python 3.12: all supported versions & Python 3.13: Django >= 4.1
+    py312-django{40,41}-drf{313,314,315},
+    py313-django41-drf{313,314,315},
+    py31{2,3}-django{42,50}-drf{314,315},
+    py31{2,3}-django{51,52}-drf315,
+    py31{2,3}-prerelease,
```
terencehonles, line 18: This env list is likely overkill with the simple `tox -l | grep $PYTHON_VERSION` selection that's used for the CI testing, so either we may want to limit the number of Python versions that are part of the CI matrix or we may want to limit the test suite, but I was enumerating all valid versions to start that discussion.

JoelLefkowitz:
Thanks @terencehonles, it's crazy to me that python3.9 is already approaching it's end of life. I'm not familiar with the `Python :: 3 :: Only` classifier, does it mean just 'not python 2'?

Do you know how the PyPI search works, or where the documentation is? If we have `Python :: 3.9` will it be included if someone searches for `Python :: 3`?

The python3.6 test run can't be removed from the CI test matrix because I don't have access to the repository settings to disable the branch protection rule that requires it. 

Only Cristi the original author has access to those settings. He's given me write access to the repository and PyPI permissions to publish which is exactly what we need to maintain the package and keep publishing new versions #662. I don't mind leaving the python3.6 run in the test matrix since it's immediately skipped.

terencehonles:
> Thanks @terencehonles, it's crazy to me that python3.9 is already approaching it's end of life. I'm not familiar with the `Python :: 3 :: Only` classifier, does it mean just 'not python 2'?

Yeah, it's crazy to me too :exploding_head:, and yes, it just means the package doesn't support Python 2. It's not really required, but it's just to be clear to any scripts that it does not support Python 2 and not just that we've omitted a Python 2 classifier. 

> Do you know how the PyPI search works, or where the documentation is? If we have `Python :: 3.9` will it be included if someone searches for `Python :: 3`?

I would kinda hope that it would do some basic substitution like this, but I really don't know, and it makes sense to just include `Python :: 3` just in case.

> The python3.6 test run can't be removed from the CI test matrix because I don't have access to the repository settings to disable the branch protection rule that requires it.
> 
> Only Cristi the original author has access to those settings. He's given me write access to the repository and PyPI permissions to publish which is exactly what we need to maintain the package and keep publishing new versions 🚀 #662. It makes sense for him to retain some of the settings access since he spent a long time developing the package from scratch and it's a big deal to give full access to someone else. I don't mind leaving the python3.6 run in the test matrix since it's immediately skipped.

I'm not sure how easy it is to contact Cristi (I'm slightly surprised you managed to do it, since I also reached out to him a bunch way back when), but does it make sense to see if he's willing to give you more permissions? The admin permission would work, but I think you're saying that might be higher than he want(ed/s) to grant, but looking at [the docs](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/managing-a-branch-protection-rule), a custom role would also suffice. I'm not sure if that's something that is available for public repos or if that's a paid feature.

terencehonles:
Looks like custom roles, are only for organizations, so the admin role would need to be granted to manage branch protections or the newer rulesets. Instead of just faking this, can you ask Cristi about getting that permission?

It would be good to enforce that the newer Python tests pass, since only `3.{6..9}` are required, and that means we need to keep all of those and just have them bail out, but the newer tests aren't actually required and can be merged failing.

JoelLefkowitz:
Thanks @terencehonles, I think we can make the test environments list a little cleaner by iterating the django version since the drf are linked directed to them:
 
```ini
py3{9,10,11,12   }-django40-drf{313,314,315},
py3{9,10,11,12,13}-django41-drf{313,314,315},
py3{9,10,11,12,13}-django42-drf{    314,315},
py3{  10,11,12,13}-django50-drf{    314,315},
py3{  10,11,12,13}-django51-drf{        315},
py3{  10,11,12,13}-django52-drf{        315},
```

I looked at some of the big packages on pypi and a lot of them have the `Python :: 3` and `Python :: 3 :: Only` flag so I agree we should add them. I would prefer if we listed the classifiers statically in the `pyproject.toml` file rather than generating them dynamically in the `setup.py` script.

As far as admin permissions go I don't think we need to ask Cristi for a custom role since we have what we need to keep publishing new versions. I appreciate that it's not pretty to bail out some test runs on every PR but I've seen from experience that when someone takes a break it's important to be extra respectful of their space.

terencehonles:
> Thanks @terencehonles, I think we can make the test environments list a little cleaner by iterating the django version since the drf are linked directed to them:
> 
> ```ini
> py3{9,10,11,12   }-django40-drf{313,314,315},
> py3{9,10,11,12,13}-django41-drf{313,314,315},
> py3{9,10,11,12,13}-django42-drf{    314,315},
> py3{  10,11,12,13}-django50-drf{    314,315},
> py3{  10,11,12,13}-django51-drf{        315},
> py3{  10,11,12,13}-django52-drf{        315},
> ```

I grouped by Python version because I thought it was easier to see, and would easily allow removing Python 3.9 when it hit EoL, but the way you put it seems reasonable since you could still remove the 9s from the list, and newer Pythons won't appear earlier in the support matrix so I went with your suggestion (putting Django 5.1 & 5.2 together though)

> I looked at some of the big packages on pypi and a lot of them have the `Python :: 3` and `Python :: 3 :: Only` flag so I agree we should add them. I would prefer if we listed the classifiers statically in the `pyproject.toml` file rather than generating them dynamically in the `setup.py` script.

Does it work to put these things partially in `pyproject.toml` and partially in `setup.py`? I would assume we want to just see about moving everything there rather than doing this change as part of this PR. It looks like we were trying to make sure our docs stayed correct by pulling the versions out of the README.

> As far as admin permissions go I don't think we need to ask Cristi for a custom role since we have what we need to keep publishing new versions. I appreciate that it's not pretty to bail out some test runs on every PR but I've seen from experience that when someone takes a break it's important to be extra respectful of their space.

I don't think we want to bail out on "some tests" because it's actually *all* tests. I came up with another solution which is a bit of a hack and lies that we are testing on older Python versions (if someone were to look at the tests on a PR), but it addresses your concern about not bugging Cristi so you can tell me what you think...

JoelLefkowitz:
Thanks @terencehonles, it's a tricky one to know how much to move to the `pyproject.toml` but my intuition is that we should move as much as possible. I'd like to move away from dynamic behaviour when packaging since it creates an opportunity for something to fail. Having a `setup.py` that does complex things, such as ours, seems to create a barrier to entry for other developers by being non standard.

In #929 I updated the runners to `ubuntu-latest` because our CI was failing. Since `python3.6` is not available for the `actions/setup-python@v5` step on the newer runners I have moved it to another job in the same action that emits a successful status like you have in this PR.

terencehonles:
> Thanks @terencehonles, it's a tricky one to know how much to move to the `pyproject.toml` but my intuition is that we should move as much as possible. I'd like to move away from dynamic behaviour when packaging since it creates an opportunity for something to fail. Having a `setup.py` that does complex things, such as ours, seems to create a barrier to entry for other developers by being non standard.
> 
> In #929 I updated the runners to `ubuntu-latest` because our CI was failing. Since `python3.6` is not available for the `actions/setup-python@v5` step on the newer runners I have moved it to another job in the same action that emits a successful status like you have in this PR.

@JoelLefkowitz I would like to get this merged instead of having the goal posts continue to move. Changes like these _can_ come in subsequent PRs, and you know what you're looking for so you can propose a PR which meets your expectations.

Having that said, I've moved the properties from `setup.py` to `pyproject.toml` as you requested. I also verified that I could move all the dependencies to the `pyproject.toml` and did so.

JoelLefkowitz (approved):

Issue #940, opened 2025-10-16
Upgrading from 1.21.7 to 1.21.11 appears to break anything that uses `TYPE_CHECKING`


beaugunderson:
# Bug Report

## Description

`drf-yasg` is failing with `name '<Name>' is not defined.`

When looking for `<Name>`, it's always a type that's conditionally imported, e.g.:

```python
if TYPE_CHECKING:
    from api.models.medication import Medication
```

And then `Medication` is used later in the file, either quoted (`"Medication"`) or unquoted, if `from __future__ import annotations` has been specified in the file.

## Is this a regression?

Yes, it worked in 1.21.7 and not in 1.21.11. I can go through the interstitial versions to pin down the version that broke it.

## Minimal Reproduction

To come, but adding a single conditional import anywhere in Django and using it should trigger it.

## Stack trace / Error message

```code
Traceback (most recent call last):
  File "/app/./manage.py", line 88, in <module>
    execute_from_command_line(sys.argv)
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^
  File "/venvs/app/lib/python3.13/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
    utility.execute()
    ~~~~~~~~~~~~~~~^^
  File "/venvs/app/lib/python3.13/site-packages/django/core/management/__init__.py", line 436, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^
  File "/venvs/app/lib/python3.13/site-packages/django/core/management/base.py", line 412, in run_from_argv
    self.execute(*args, **cmd_options)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^
  File "/venvs/app/lib/python3.13/site-packages/django/core/management/base.py", line 458, in execute
    output = self.handle(*args, **options)
  File "/venvs/app/lib/python3.13/site-packages/drf_yasg/management/commands/generate_swagger.py", line 194, in handle
    schema = self.get_schema(generator, request, not private)
  File "/venvs/app/lib/python3.13/site-packages/drf_yasg/management/commands/generate_swagger.py", line 136, in get_schema
    return generator.get_schema(request=request, public=public)
[...]

PR #941, opened 2025-10-20
Handle annotations not available at runtime

JoelLefkowitz:
As shown in #940, when a type is not available at runtime `typing.get_type_hints` will throw an error. This introduced a bug from `v1.21.10` onwards.

This PR introduces a fallback by wrapping the `typing.get_type_hints` call in a try/except block and reverting to the raw annotation when the type resolution fails.


beaugunderson:
I tested with this branch and it does solve our issue! Thank you!

PR #945, opened 2025-12-23
Update logout button to use POST request.

### LINKED ISSUE #733: setting "LOGOUT_URL" has no effect!
Hi,
As documented [here](https://drf-yasg.readthedocs.io/en/stable/settings.html#logout-url), setting variable `LOGOUT_URL` in `settings.py` should override the default value of `LOGOUT_URL` (which sets to `/accounts/logout/` by default).
but it seems it has no effect! regardless of setting `LOGOUT_URL` or not, the `LOGOUT_URL` will always be `/accounts/logout/`.

m000:
Presumably, most users use a LOGOUT_URL that maps to Django's LogoutView. However, using GET with LogoutView has been deprecated since Django 4.1. We update the logout button to use POST instead.

This may be a breaking change for any users that have implemented a custom GET-based logout page. These users can either switch to LogoutView which now works again out of the box. Or, adapt their view to also accept POST requests. Or, override `swagger-ui.html` template to use a simple `<a>` link.

Fixes #733.

See: https://docs.djangoproject.com/en/dev/releases/4.1/#log-out-via-get

### DIFF
--- a/src/drf_yasg/static/drf-yasg/style.css
+++ b/src/drf_yasg/static/drf-yasg/style.css
@@ -71,3 +71,12 @@ svg.swagger-defs {
     width: 0;
     height: 0;
 }
+
+button.header__btn {
+    background: none;
+    border: none;
+    cursor: pointer;
+    padding: 0;
+    font: inherit;
+    color: inherit;
+}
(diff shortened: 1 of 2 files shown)


m000:
Maybe some documentation update would also be needed. I will follow with these updates as soon as we finalize the code part.

JoelLefkowitz:
Thanks @m000 looks good!

PR #964, opened 2026-07-02
Simplify zoneinfo import

eagleoflqj:
Since [zoneinfo](https://docs.python.org/3/library/zoneinfo.html) is added to standard lib in 3.9 and 3.10+ is required here, it's no longer necessary to try importing it.


JoelLefkowitz:
Thanks @eagleoflqj

